PR #19, opened 2018-08-15
Slim down build config

mikkeloscar:
* golint installed by default in overlay
* `GOPATH` set by default in overlay
* `GOBIN` set by default in overlay


mikkeloscar:
:+1:

PR #29, opened 2018-08-22
Add support for custom service-ports

### LINKED ISSUE #9: Add support for backend and service ports
Right now we assume `serviceport==targetPort`, but it should be optionally configurable

mikkeloscar:
This adds support for specifying custom service ports if desired.

The basic idea is that you can specify a `service` key in the stack template of the `stackset` and then set custom ports:

```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: my-app-svc
spec:
  ingress:
    hosts: [my-app.example.org, alt.name.org]
    backendPort: 80 # if this is not set, it defaults to `ingress` (old behavior)
  stackLifecycle:
    scaledownTTLSeconds: 300
    limit: 5
  stackTemplate:
    spec:
      # optional service.ports definitions.
      # If not set you get a single port that matches the ingress backendPort and a pod port.
      service: 
        ports:
        - name: 'main'
          port: 80
          targetPort: 80
        - name: 'metric'
          port: 90
          targetPort: 80
      version: v1
      replicas: 3
      horizontalPodAutoscaler:
        minReplicas: 3
        maxReplicas: 10
        metrics:
        - type: Resource
          resource:
            name: cpu
            targetAverageUtilization: 50
      podTemplate:
        spec:
          containers:
          - name: nginx
            image: nginx
            ports:
            # port should match ingress.backendPort either by name or port number.
            # if it doesn't match no service will be created.
            - containerPort: 80
              name: ingress
            resources:
              limits:
                cpu: 10m
                memory: 50Mi
              requests:
                cpu: 10m
                memory: 50Mi
```

### TODO
* [x] Document in more detail how this works.

Fix #9


docs/howtos.md
```diff
@@ -0,0 +1,129 @@
+# How to's
+
+## Configure port mapping
+
+`StackSets` creates the underlying resource types `Ingress`, `Service` and
```
szuecs: Do you want to hide the existence of Stack?
StackSets -> Stack -> Ingress,Svc,..

docs/howtos.md
```diff
@@ -0,0 +1,129 @@
...
+```yaml
+containers:
+- name: nginx
+  image: nginx
+  ports:
+  - containerPort: 80
+    name: ingress
+```
+
+The service created for a `Stack` will get the following generated port
```
szuecs: Here you wrote about `Stack` which was not introduced

docs/howtos.md
```diff
@@ -0,0 +1,129 @@
...
+  port: 80
+  protocol: TCP
+  targetPort: 80
+```
+
+Which ensures that there is a connection from `Ingress -> Service -> Pods`.
+
+If you have multiple ports or containers defined in a pod it's important that
+exactly one of the ports map to the `backendPort` defined for the ingress
+(default value: `ingress`).
```
szuecs: What happens if you don't do that (exactly one port)?

docs/howtos.md
```diff
@@ -0,0 +1,129 @@
...
+            image: nginx-with-metrics
+            ports:
+            - containerPort: 8080
+              name: ingress
+            - containerPort: 7979
+              name: metrics
+```
+
+Here you must make sure that `spec.ingress.backendPort` map to one of the ports
+defined for the service and that the service ports map to container ports.
```
szuecs: How you make it sure?

controller/stack.go
```diff
@@ -597,10 +600,36 @@ func getStackService(kube kubernetes.Interface, deployment appsv1.Deployment) (*
 	return service, nil
 }
 
+// getServicePorts gets the service ports to be used for the stack service.
+func getServicePorts(backendPort intstr.IntOrString, stack zv1.Stack) ([]v1.ServicePort, error) {
+	var servicePorts []v1.ServicePort
+	if stack.Spec.Service == nil || len(stack.Spec.Service.Ports) == 0 {
+		servicePorts = servicePortsFromContainers(stack.Spec.PodTemplate.Spec.Containers)
+	} else if stack.Spec.Service != nil {
```
szuecs: I think the condition is wrong and you can use `else` without `if`

mikkeloscar: I have to check that `stack.Spec.Service != nil` otherwise it will crash on the next line :)

mikkeloscar:
@szuecs I updated the docs with your suggestions, PTAL.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #30, opened 2018-08-22
Make it possible to run multiple controllers

### LINKED ISSUE #28: Make it possible to mark stacksets owned by a certain controller.
When testing changes to the stackset-controller it's hard to do so in a cluster where it's already deployed because it will own ALL stacks in the cluster.

We should have a concept of ownership where an annotation on the stackset would let the controller know if it owns it or not. Similar to the ingress class annotation.

mikkeloscar:
This makes it possible to run multiple controllers in a single cluster.

This is possible by defining a `controller-id` for the stackset-controller which means it will only manage `StackSets` where the annotation `stackset-controller.zalando.org/controller` is defined with a value matching the `controller-id`.

Fix #28

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -109,9 +109,8 @@ func (c *StackSetController) Run(ctx context.Context) {
 				continue
 			}
 
-			// check if stackset should be managed by the
-			// controller
-			if !c.shouldManage(&stackset) {
+			// check if stackset should be managed by the controller
+			if !c.hasOwnership(&stackset) {
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


controller/stackset.go
```diff
@@ -106,6 +109,12 @@ func (c *StackSetController) Run(ctx context.Context) {
 				continue
 			}
 
+			// check if stackset should be managed by the
+			// controller
```
szuecs: this can be in the row before

controller/stackset.go
```diff
@@ -153,6 +162,20 @@ func (c *StackSetController) Run(ctx context.Context) {
 	}
 }
 
+// shouldManage returns true if the controller should manage the stackset.
+// Whether it should manage is determined by the value of the
+// 'stackset-controller.zalando.org/controller' annotation. If the value
+// matches the controllerID then it should manage it, or if the controllerID is
+// "" and there's no annotation set.
+func (c *StackSetController) shouldManage(stackset *zv1.StackSet) bool {
```
szuecs: maybe a btter name would be `hasOwnership()`

szuecs:
:+1:

mikkeloscar:
:+1:

Issue #3, opened 2018-08-10
 Deletion of StackSet doesn't cleanup Stacks and Deployments


szuecs:
`kubectl delete stackset <foo>` removes the `StackSet` but doesn't cleanup the `Stack` objects.

Once the `finalizer` is removed manually it gets cleanup up. So I guess, the deletion handler currently just doesn't process the finalizers.

mikkeloscar:
Fixed in #32

Issue #2, opened 2018-08-10
"Force" delete Stack objects 


szuecs:
`kubectl delete stack <foo>` seems to hang and never deletes the stack.

This seems to be by design due to the `finalizer`. Let's deal with it later.

mikkeloscar:
This is intended to ensure you don't delete stacks which are getting traffic. There is one edge case where you can't delete the last stack when it's the only one getting traffic, maybe we need to allow "force" deleting of the stack if it's desired.

mikkeloscar:
"Fixed" in #32

PR #32, opened 2018-08-24
Drop finalizers and use ownership model from Kubernetes

### LINKED ISSUE #27: Latest stack without traffic can't be deleted
When I deploy two versions of an app it is not possible to delete the latest stack even though it has no traffic. If it is in a crash loop or smoke tests failed I would like to get rid of that stack. Currently the stack is being recreated. This behavior might be surprising and confusing. 

Should we adopt this? Any thoughts?

mikkeloscar:
This simplifies the ownership model of `StackSet` -> `Stack` -> `Deployment` -> `Service`. to use the ownership model built into Kubernetes.

Before I tried to use finalizers to "stop" termination of a stack in case it was still getting traffic. But there were many unfixed edge cases in the implementation and in the end it's a hack to try and block deletion using finalizers because once you delete in Kubernetes the deletion is started and you can not undo it again.

For the user this means, that if they delete a stack, whether it gets traffic or not it will just get deleted straight away including deployment and service resources owned by the stack. Similar, if you delete a `StackSet` all stacks of that stackset will be deleted.

If we want to prevent users from deleting wrong stacks accidentially I suggest we implement this client side (at least for now) where it's much easier to prevent issuing the delete of the resource in the first place instead of trying to stop the already started deletion via a finalizer.

Closes #27, #3, #2

### DIFF
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

Issue #8, opened 2018-08-10
Add support to stackLifecycle


szuecs:
Implement the field `stackLifecycle`

```
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: my-app
spec:
  # optional Ingress definition.
  ingress:
    hosts: [my-app.example.org, alt.name.org]
  stackLifecycle:
    scaledownTTLSeconds: 300
    limit: 5
  stackTemplate:
   ....
```

stackLifecycle definition tells when stackset-controller should clean up the resources from and old stack that gets no traffic anymore.

- scaledownTTLSeconds: number of seconds to wait before scaling down the replicas of a stack to 0
- limit: at most N stacks should have no traffic

linki:
@szuecs How should this work?

szuecs:
I ported the issues, but I guess @mikkeloscar knows more :)

mikkeloscar:
Fixed in #33

PR #33, opened 2018-08-24
#8 support stackLifecycle to make autocleanup work

szuecs:
#8 support stackLifecycle to make autocleanup work

Instead of using a fixed scaledownTTLSeconds for all stackset get the value if defined from the stackLifecycle spec. The `limit`was already used correctly.

### DIFF
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -253,8 +253,8 @@ func (c *StackController) manageDeployment(stack zv1.Stack, traffic map[string]T
 						)
 					}
 				}
+				return nil
 			}
-			return nil
 
 		} else {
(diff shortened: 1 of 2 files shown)


controller/stack.go
```diff
@@ -255,8 +253,9 @@ func (c *StackController) manageDeployment(stack zv1.Stack, traffic map[string]T
 						)
 					}
 				}
-				return nil
 			}
+			return nil
```
mikkeloscar: ups, this looks wrong now

szuecs:
:+1:

mikkeloscar:
:+1:

PR #35, opened 2018-08-26
refactor: Improve controller performance

mikkeloscar:
This refactors the controller code to improve the performance and
optimize the number of calls made to the API server.

In the original implementation it created several goroutines per
`StackSet`. Each goroutine would poll the same resources resulting in a
lot of unessesary Get calls to the API server. The number of calls
increased with the number of `StackSets` managed by the controller.
The following calls were made:

    n x [Get Endpoints]
    n x [Get Ingress] x 3
    n x [List Stacks] x 5
    n x [Get Deployment]
    n x [Get Service]
    n x [Get HPA]

Where `n` is the number of `StackSets`. This sums up to `12 x n` `Get`
calls for each iteration of the control loops.

The new implementation fetches all the resources in the main loop and
organizes them in `StackSetContainers` which is a struct storing all the
resources related to a single StackSet. The `StackContainers` are then
passed to individual reconciler loops which works the same as before
except the read from the in memory representation instead of fetching
reasources from the API server. This reduces the number of `Get` calls
from `12 x n` to `6` for each iteration of the control loop.

### DIFF
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -231,8 +231,15 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 	}
 
 	if !reflect.DeepEqual(newStatus, stack.Status) {
+		c.logger.Infof(
+			"Status changed for Stack %s/%s: %#v -> %#v",
+			stack.Namespace,
+			stack.Name,
+			stack.Status,
+			newStatus,
+		)
 		stack.Status = newStatus
-		// TODO: log the change in status
+
 		// update status of stack
 		_, err = c.appClient.ZalandoV1().Stacks(stack.Namespace).UpdateStatus(&stack)
 		if err != nil {
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -499,6 +499,40 @@ func applyPodTemplateSpecDefaults(template v1.PodTemplateSpec) v1.PodTemplateSpe
 		if container.ImagePullPolicy == "" {
 			newTemplate.Spec.Containers[i].ImagePullPolicy = v1.PullIfNotPresent
 		}
+		if container.ReadinessProbe != nil {
+			if container.ReadinessProbe.Handler.HTTPGet != nil && container.ReadinessProbe.Handler.HTTPGet.Scheme == "" {
+				newTemplate.Spec.Containers[i].ReadinessProbe.Handler.HTTPGet.Scheme = v1.URISchemeHTTP
+			}
+			if container.ReadinessProbe.TimeoutSeconds == 0 {
+				newTemplate.Spec.Containers[i].ReadinessProbe.TimeoutSeconds = 1
+			}
+			if container.ReadinessProbe.PeriodSeconds == 0 {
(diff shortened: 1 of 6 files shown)


szuecs:
Can you add godoc for new core structs, please?
It will be easier to understand.
The reduction of API calls are easy to spot, but I miss the documentation of the "data idea". For example ssc.Ingress vs ssc.Stack.Spec.Ingress.

mikkeloscar:
@szuecs added some struct documentation. Hope it's more clear now, otherwise let me know :)

controller/stack.go
```diff
@@ -311,57 +204,49 @@ func (c *StackController) manageDeployment(stack zv1.Stack, traffic map[string]T
...
 	}
 
-	// TODO: log the change in status
-	// update status of stackset
-	_, err = c.appClient.ZalandoV1().Stacks(stack.Namespace).UpdateStatus(&stack)
-	if err != nil {
-		return err
+	if !reflect.DeepEqual(newStatus, stack.Status) {
+		stack.Status = newStatus
+		// TODO: log the change in status
```
szuecs: should this reference an issue?

controller/stackset.go
```diff
@@ -125,51 +180,281 @@ func (c *StackSetController) Run(ctx context.Context) {
...
+// StackSetContainer is a container for storing the full state of a StackSet
+// including the sub-resources which are part of the StackSet. It respresents a
+// snapshot of the resources currently in the Cluster. This includes an
+// optional Ingress resource as well as the current Traffic distribution. It
+// also contains a set of StackContainers which respresents the full state of
+// the individual Stacks part of the StackSet.
+type StackSetContainer struct {
+	StackSet        zv1.StackSet
+	StackContainers map[types.UID]*StackContainer
+	Ingress         *v1beta1.Ingress
```
szuecs: This Ingress is the main ingress that has the full traffic distribution, right?
StackSet has also ingresses, which are the versioned ingresses, is that correct?
If so, this would be nice to add for the Ingress attribute of the StackSetContainer

controller/stackset.go
```diff
@@ -125,51 +180,281 @@ func (c *StackSetController) Run(ctx context.Context) {
...
+	Traffic         map[string]TrafficStatus
```
szuecs: please add doc: Traffic is the current traffic distribution across stacks (on ingress?)

what is the key "string", a service, a stack or an ingress?

controller/stackset.go
mikkeloscar: > This Ingress is the main ingress that has the full traffic distribution, right?

Yes

> StackSet has also ingresses, which are the versioned ingresses, is that correct?

That's true, these are actually not part of this whole data-structure, but they probably should be. I would like to defer this to a separate PR if you don't mind?

I added some documentation about the `Ingress` field.

controller/stack.go
mikkeloscar: Replaced the TODO with the logs :)

controller/stackset.go
mikkeloscar: Good point! It's the stack name. I added a comment about this.

mikkeloscar:
:+1:

PR #39, opened 2018-08-28
Unify multiple clients into one

linki:
Unifies both core Kubernetes as well as Zalando's Stackset client into one.

Not sure if it's good but it simplifies the setup a bit since we don't have to deal with multiple clients.

### DIFF
--- a/cmd/traffic/main.go
+++ b/cmd/traffic/main.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/alecthomas/kingpin"
 	log "github.com/sirupsen/logrus"
+	"github.com/zalando-incubator/stackset-controller/pkg/clientset"
 	"github.com/zalando-incubator/stackset-controller/pkg/traffic"
 	rest "k8s.io/client-go/rest"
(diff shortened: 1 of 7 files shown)


pkg/traffic/clientset.go
```diff
@@ -0,0 +1,41 @@
+package traffic
```
mikkeloscar: What about making it even more generic so it's all of `kubernetes.Interface` + `ZalandoV1()`? and having it outside the traffic package. Then it could be reused in the controller code (doesn't have to be done in the same PR).

linki: That was my goal but apparently forgot about it 🤷‍♂️  Will do

linki: I would still just add we need at this point. wdyt?

mikkeloscar: yes, that's fine.

linki: done 😄

linki:
👍

PR #41, opened 2018-09-02
Use equality.Semantic.DeepEqual for resource comparison

mikkeloscar:
Use `equality.Semantic.DeepEqual` instead of `reflect.DeepEqual` for correct comparison of special Kubernetes types like `resource.Quantity`.

The `resource.Quantity` type can represent the same value in different representations i.e. `2000m` is equal to `2`. In that case the internal fields of the struct will look differently and therefore it's not possible to compare just with `reflect.DeepEqual`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -53,33 +53,6 @@
   version = "v1.0.0"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


szuecs:
👍

mikkeloscar:
:+1:

PR #51, opened 2018-09-06
Use stackLifecycle config to determine stacks to be garbage collected

mikkeloscar:
Use stackLifecycle config to determine stacks to be garbage collected


njuettner:
👍

mikkeloscar:
:+1:

Issue #26, opened 2018-08-21
Create canonical Service object


linki:
https://github.bus.zalan.do/teapot/cluster-registry-deploy/pull/39

stackset controller manages the Service object, so users *shouldn't* mess with it. However, currently it creates a Service for each version with a corresponding name. With this, cluster-local client applications that intend to use the Service's DNS name don't have a stable access point anymore because they always include the version.

Let's also setup a unversioned Service object that points to some application pods. They can't be traffic switched so the question is, what should be the selector?

mikkeloscar:
Good idea!

>  They can't be traffic switched so the question is, what should be the selector?

`stackset: <name>` could be the selector. [Each stack has this](https://github.com/zalando-incubator/stackset-controller/blob/4fc7a1fd1b0524fc029b7533ad57ab63a5fbfa6c/controller/stackset.go#L473).

linki:
> stackset: <name> could be the selector.

Right, but that would mean 100% on a particular version. Wrong if you have a 20/80% split on the Ingress level but probably good enough.

muaazsaleem:
It seems there are some questions around this feature:

**What are the use cases for it?**
I only see one use case for this. If someone is switching from using a normal service to `StackSet`, their application would go from having a stable `cluster.local` endpoint  like
`some-service.default.svc.cluster.local` to only having per version service endpoints i.e `some-service-v*.default.svc.cluster.local`. 

This might break callers that were using the original `cluster.local` endpoint. This is specially a common use case for UI/API Gateways e.g [tailor](https://github.com/zalando/tailor) or [zuul](https://github.com/Netflix/zuul) because they usually refer to components using the `cluster.local` endpoints.

**Which Stack version would this service point to?**
Since we can't _really_ traffic switch on service endpoints, which Stack version should this service point to. Should it always point to the latest version or should that be configurable?

One solution could be that we have a `service` key in the spec that refers to a stack version, something like:
```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: my-app
spec:
  #... optional igress definition
  # define which Stack service should the cluster.local endpoint point to 
 service:
    version: v1 # direct 100% of the cluster.local traffic to this version of the Stack.
```

But this would mean that essentially traffic switching is "broken" for traffic coming from inside the cluster. Besides, people could also achieve the same effect by defining a service explicity, so what’s the additional benefit of us implementing this?

linki:
[...]

PR #54, opened 2018-09-14
Adding the first draft for a per StackSet service

muaazsaleem:
Service endpoints are typically stable in k8s. StackSet controller changes
that because while traffic switching a StackSet may have different Services
that point to different versions of an application.

Trying to add a per StackSet service, so it's endpoint would serve as a stable
endpoint to the application

ref: https://github.com/zalando-incubator/stackset-controller/issues/26


linki:
@muaazsaleem @mikkeloscar We decided to not implement this feature. Can this be closed then?

mikkeloscar:
> Can this be closed then?

Fine by me.

linki:
We decided to not implement this feature.

Issue #12, opened 2018-08-10
Send relevant events to Stack/StackSet resources.


szuecs:
The controller should send relevant events to the `Stack` and `StackSet` resources so users know what it's doing.

PR #55, opened 2018-09-14
Event recorder

njuettner:
* send custom events to Kubernetes to be recorded for targeted Kubernetes objects

Fix for #12

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -14,10 +14,12 @@ import (
 	log "github.com/sirupsen/logrus"
 	zv1 "github.com/zalando-incubator/stackset-controller/pkg/apis/zalando/v1"
 	"github.com/zalando-incubator/stackset-controller/pkg/clientset"
+	"github.com/zalando-incubator/stackset-controller/pkg/recorder"
 	"golang.org/x/sync/errgroup"
 	appsv1 "k8s.io/api/apps/v1"
 	autoscaling "k8s.io/api/autoscaling/v2beta1"
 	"k8s.io/api/core/v1"
+	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


controller/stackset.go
```diff
@@ -157,7 +163,7 @@ func (c *StackSetController) Run(ctx context.Context) {
 				continue
 			}
 
-			c.logger.Infof("Adding entry for StackSet %s/%s", stackset.Namespace, stackset.Name)
+			c.recorder.Eventf(e.StackSet, apiv1.EventTypeNormal, "CreateStackSet", "StackSet '%s/%s' added", stackset.Namespace, stackset.Name)
```
mikkeloscar: This is more internal logging for the controller. I think it's a bit confusing to log as an event as it will happen every time the controller restarts.

pkg/clientset/unified.go
```diff
@@ -3,31 +3,115 @@ package clientset
...
+	Storage() storagev1.StorageV1Interface
+	StorageV1alpha1() storagev1alpha1.StorageV1alpha1Interface
 	ZalandoV1() zalandov1.ZalandoV1Interface
 }
 
 type Clientset struct {
-	kubernetes kubernetes.Interface
-	stackset   stackset.Interface
+	kubernetes.Interface
+	stackset stackset.Interface
```
linki: can we drop `stackset` as well?

mikkeloscar: How do you mean? We can't do:

```go
type x struct {
  kubernetes.Interface
  stackset.Interface
}
```

as the two `Interface` field names would overlap.

linki: That's what I meant, thanks.

linki:
:+1:

njuettner:
👍

PR #59, opened 2018-09-19
Discard logs from glog

njuettner:
* discard noisy logs like ```code ERROR: logging before flag.Parse: E0918 10:56:44.231313       1 streamwatcher.go:109] Unable to decode an event from the watch stream: net/http: request canceled (Client.Timeout exceeded while reading body)```


linki:
This also allows for a read-only container fs 💃

linki:
:+1:

linki:
@szuecs can you have a look as well?

njuettner:
The replaced glog package which we use is already initialising a logger which discards all logs, so we don't need to initialise a nullLogger.
```code
var logger logrus.FieldLogger = &logrus.Logger{
	Out:   ioutil.Discard
...
```

linki:
:+1:

PR #61, opened 2018-09-20
reconcile should not create a new event recorder

njuettner:
stackset-controller crashes due to OOM. because of initialising eventrecorder in each reconcile loop

screen shot 2018-09-20 at 14 39 35

screen shot 2018-09-20 at 14 39 01


mikkeloscar:
:+1:

linki:
:+1:

PR #68, opened 2018-10-04
Validate targetValue quantity values

mikkeloscar:
Validate `targetValue` and `targetAverageValue` values as either string
or integer. This allows defining HPAs like this:

```yaml
horizontalPodAutoscaler:
  minReplicas: 3
  maxReplicas: 10
  metrics:
  - type: Object
    object:
      metricName: requests-per-second
      target:
        apiVersion: extensions/v1beta1
        kind: Ingress
        name: my-app
      targetValue: 100 # this would not be allowed before
```

This is similar to the fix done in #57

### DIFF
--- a/docs/stack_crd.yaml
+++ b/docs/stack_crd.yaml
@@ -93,9 +93,15 @@ spec:
                                   type: string
                                 type: object
                           targetAverageValue:
-                            type: string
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

PR #70, opened 2018-10-10
Allow stacks to play nice with kube-downscaler.

arjunrn:
When the `replicas` on the stack is set to 0 then the deployment should be scaled down regardless of whether it's receiving traffic or not. If it's scaled down and the replicas in stack are set to a non-zero value then it should be scaled back up. There is an override to set the replica count to 0 if the stack is not receiving traffic later in the code.

### DIFF
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -139,13 +139,13 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 	}
 
 	// If the deployment is scaled down by the downscaler then scale it back up again
-	if *deployment.Spec.Replicas == 0 && *stack.Spec.Replicas != 0 {
+	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas == 0 && *stack.Spec.Replicas != 0 {
 		replicas := int32(*stack.Spec.Replicas)
 		deployment.Spec.Replicas = &replicas
 	}
 
 	// The deployment has to be scaled down because the stack has been scaled down then set the replica count
-	if *deployment.Spec.Replicas != 0 && *stack.Spec.Replicas == 0 {
+	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != 0 && *stack.Spec.Replicas == 0 {
 		replicas := int32(0)
 		deployment.Spec.Replicas = &replicas
 	}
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

controller/stack.go
```diff
@@ -138,6 +138,18 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 		deployment.Spec.Replicas = stack.Spec.Replicas
 	}
 
+	// If the deployment is scaled down by the downscaler then scale it back up again
+	if *deployment.Spec.Replicas == 0 && *stack.Spec.Replicas != 0 {
```
mikkeloscar: I think we need to check that `Replicas != nil` here.

controller/stack.go
```diff
@@ -138,6 +138,18 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 		deployment.Spec.Replicas = stack.Spec.Replicas
 	}
 
+	// If the deployment is scaled down by the downscaler then scale it back up again
+	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas == 0 && *stack.Spec.Replicas != 0 {
```
mikkeloscar: Same for `*stack.Spec.Replicas` here it could also be `nil`.

controller/stack.go
```diff
@@ -138,6 +138,22 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 		deployment.Spec.Replicas = stack.Spec.Replicas
 	}
 
+	// If the deployment is scaled down by the downscaler then scale it back up again
+	if stack.Spec.Replicas != nil && *stack.Spec.Replicas != 0 {
+		if deployment.Spec.Replicas == nil || *deployment.Spec.Replicas == 0 {
+			replicas := int32(*stack.Spec.Replicas)
```
mikkeloscar: This should only be done if autoscaling is not enabled, otherwise it would reset the value all the time. (sorry for not seeing this in the first review).

mikkeloscar:
:+1:

controller/stack.go
arjunrn: If autoscaling is enabled then the condition `deployment.Spec.Replicas == nil || *deployment.Spec.Replicas == 0` will not evaluate to `true` if the replicas field is populated correctly.

arjunrn: > nvm. It only sets it if it was already 0 so this wouldn't be an issue

Which is what you would expect to happen when the downtime is over.

arjunrn:
:+1:

PR #71, opened 2018-10-17
Don't downscale the current stack even if it has no traffic

aermakov-zalando:
This allows the deployment to proceed even if the users set a short TTL. A downside is that they'll have a deployment which doesn't do anything if they decide to go back to a previous version, but it looks like an acceptable trade-off.


mikkeloscar:
Seems like a reasonable trade-off to me.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #73, opened 2018-10-22
Implemented the simplified autoscaling spec for stacksets.

arjunrn:
This PR adds a simplified spec for declaring HPAs. The following is an example of the format:
```yaml
kind: StackSet
spec:
  autoscaling:
    minReplicas: 1
    maxReplicas: 3
    metrics:
    - type: PodJson
      endpoint:
        port: 9090
        path: /metrics
        key: '$.http_server.rps'
      average: 1k
    - type: AmazonSQS
      queue:
        name: foo
        region: eu-west-1
      average: 30
    - type: Ingress
      ingress:
        name: my-app
      average: 30
    # Normal metrics are supported as well
    - type: CPU
      averageUtilization: 80
    - type: Memory
      averageUtilization: 80
```

In it's current behavior while reconciling the stacksets if the `autoscaler` is specified and the `horizontalPodAutoscaler` is not specified then a hpa object is generated. If the hpa is already specified then the autoscaler is ignored.

Signed-off-by: Arjun Naik <arjun.naik@zalando.de>


controller/autoscaler.go
```diff
@@ -0,0 +1,166 @@
...
+	autoscaler := container.StackSet.Spec.StackTemplate.Spec.Autoscaler
+	stacksetName := container.StackSet.Name
+	if autoscaler != nil {
+		var generatedHPA zv1.HorizontalPodAutoscaler
+		generatedHPA.MinReplicas = autoscaler.MinReplicas
+		generatedHPA.MaxReplicas = autoscaler.MaxReplicas
+		generatedHPA.Metrics = make([]autoscaling.MetricSpec, len(autoscaler.Metrics))
+
+		for i, m := range autoscaler.Metrics {
+			var generated autoscaling.MetricSpec
```
njuettner: var (
generated autoscaling.MetricSpec
err error
)

controller/stackset.go
```diff
@@ -93,10 +93,19 @@ func (c *StackSetController) Run(ctx context.Context) {
 
 			var reconcileGroup errgroup.Group
 			for stackset, container := range stackContainers {
+
+				err = c.ReconcileAutoscalers(container)
+				if err != nil {
+					c.recorder.Event(&container.StackSet, v1.EventTypeWarning,
+						"GenerateHPA", fmt.Sprintf("Failed to generate HPA %v", err.Error()))
+					c.logger.Error(err)
```
njuettner: no need to log it twice, c.recorder will also log it as well

controller/stackset.go
```diff
@@ -93,10 +93,19 @@ func (c *StackSetController) Run(ctx context.Context) {
...
+				}
+
 				container := *container
 
 				reconcileGroup.Go(func() error {
 					if _, ok := c.stacksetStore[stackset]; ok {
+
```
njuettner: omit the line

controller/stackset.go
```diff
@@ -93,10 +93,19 @@ func (c *StackSetController) Run(ctx context.Context) {
 
 			var reconcileGroup errgroup.Group
 			for stackset, container := range stackContainers {
+
```
njuettner: omit the line

controller/autoscaler.go
```diff
@@ -0,0 +1,166 @@
...
+			generatedHPA.Metrics[i] = generated
+		}
+		container.StackSet.Spec.StackTemplate.Spec.HorizontalPodAutoscaler = &generatedHPA
+	}
+	return nil
+}
+
+func CPUMetric(metrics zv1.AutoscalerMetrics) (autoscaling.MetricSpec, error) {
+	if metrics.AverageUtilization == nil {
+		return autoscaling.MetricSpec{}, fmt.Errorf("utilization is not specified")
```
njuettner: Out of curiosity:

Why didn't you return *autoscaling.MetricSpec, then you could return nil with the error.  Because I think it's more preferred to return nil in Go.

controller/stackset.go
```diff
@@ -93,10 +93,19 @@ func (c *StackSetController) Run(ctx context.Context) {
...
+				err = c.ReconcileAutoscalers(container)
+				if err != nil {
+					c.recorder.Event(&container.StackSet, v1.EventTypeWarning,
+						"GenerateHPA", fmt.Sprintf("Failed to generate HPA %v", err.Error()))
+					c.logger.Error(err)
+				}
```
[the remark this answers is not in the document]
arjunrn: I think conceptually the collecting the resources is not the same as generating the the HPA spec so I put it there, but if you think it would be better in the `collectResources` function I can move it as well.

controller/autoscaler.go
```diff
@@ -0,0 +1,166 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strconv"
+)
+
+const (
+	AmazonSQSMetricName   = "AmazonSQS"
+	PodJSONMetricName     = "PodJson"
+	IngressMetricName     = "Ingress"
+	CPUMetricName         = "CPU"
```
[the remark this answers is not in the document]
arjunrn: Added

njuettner:
👍

mikkeloscar:
Only missing thing would be documentation, but maybe you can add that later as we also need docs internally, and could probably re-use the same.

docs/howtos.md
```diff
@@ -128,3 +128,70 @@ Here you must make sure that the value used for `spec.ingress.backendPort` also
...
+```yaml
+type: PodJson
+endpoint:
+  port: 9090
+  path: /metrics
+  key: '$.http_server.rps'
+average: 1k
+```
+
+If the [kube-aws-ingress-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller) is used in
```
mikkeloscar: Skipper is the relevant component. You could run without the kube-ingress-aws-controller for this (It only creates the ALBs).

mikkeloscar: And you also need kube-metrics-adapter with prometheus support enabled for this to work.

arjunrn: I've mentioned the kube-metrics-adapter earlier in the note.

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #75, opened 2018-10-26
Fix wrong naming for json marshaling

mikkeloscar:
The wrong name `metricType` was used for the `json:""` annotation of the autoscaler type.

PR #76, opened 2018-10-26
Update type name in autoscaler validation

mikkeloscar:
Forgot to update the field name in https://github.com/zalando-incubator/stackset-controller/pull/75


mikkeloscar:
:+1:

PR #77, opened 2018-10-29
Implement stack prescale logic

mikkeloscar:
This is a rough first draft for adding stack prescale logic to the controller. I have tried to pull out the scaling related code into what I call `TrafficReconciler` and implemented a `SimpleTrafficReconciler` which uses the same logic as before and added a `PrescaleTrafficReconciler` which implements the prescale logic.

Suggestions on how to improve this are very welcome. Currently it's passing a lot of information down because a lot of information is needed to decide how to do the traffic switching.

## TODO

* [x] Add a lot more tests
* [x] Test for real and see if behavior is as expected.
* [x] Document the prescaling logic.

### DIFF
--- a/controller/prescale_reconciler.go
+++ b/controller/prescale_reconciler.go
@@ -15,19 +15,17 @@ const (
 	prescaleAnnotationKey = "stacksetstacks.zalando.org/prescale-replicas"
 )
 
-type PrescaleTrafficReconciler struct {
-	base TrafficReconciler
-}
+type PrescaleTrafficReconciler struct{}
 
 // ReconcileDeployment prescales the deployment if the prescale annotation is
 // set. Prescale annotation will only be removed from the deployment after it's
 // getting traffic.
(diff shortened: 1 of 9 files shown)


controller/prescale_reconciler.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	prescaleAnnotationKey = "stacksetstacks.zalando.org/prescale-replicas"
+)
+
+type PrescaleTrafficReconciler struct {
+	base TrafficReconciler
```
linki: Seems like `base` isn't used after all.

linki:
The more code the less comments..
Looking forward to seeing this in action.
lgtm 👍

mikkeloscar:
Added tests, fixed bugs we found during manual testing and added docs.

PTAL!

linki:
:+1:

mikkeloscar:
:+1:

PR #78, opened 2018-11-05
Use StackSet instead of Application in traffic command

mikkeloscar:
The term `Application` was still used in the `traffic` CLI. Change it to `StackSet`.


mikkeloscar:
:+1:

PR #80, opened 2018-11-14
Fix generated clients

aermakov-zalando:
* codegen: use `dep` to fetch the generator instead of manual cloning
* Dependencies: use to 1.11 instead of random hashes
* Run the code generator with the correct group name (`zalando.org`) so the fake client is actually usable


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #85, opened 2018-12-11
Keep the HPA for the current stack as well

### LINKED ISSUE #84: Traffic Switch issues with CPU and RPS scaling enabled
We have an app running around 30 pods.

When we switch traffic, we have noticed 2 issues which appeared after we added RPS scaling:

In the following example, we are running stackset rendering-engine-986 and trying to deploy stackset rendering-engine-990:

screenshot 2018-12-10 at 23 04 00

- First it starts scaling previous more than stackset 2 times, meaning we end up with 70 pods on current stackset.
- Second, when it spawns the new stackset, it seems to request a correct amount of pods (35) but then it quickly requests an insane amount of pods and we go up to 87.

Fortunately after a while everything goes back to normal but these should be investigated.

Zalando internal issue with more context: https://github.bus.zalan.do/teapot/issues/issues/1548

aermakov-zalando:
The current version of the stack is always kept since https://github.com/zalando-incubator/stackset-controller/pull/71 to avoid issues with long deployments. However, the HPA deletion code didn't account for that, so the controller just deleted the HPA. Fix by keeping the HPA if the deployment is forcibly kept.

Fix #84 (one part of).

### DIFF
--- a/controller/stack.go
+++ b/controller/stack.go
@@ -155,9 +155,10 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 	}
 
 	currentStackName := generateStackName(ssc.StackSet, currentStackVersion(ssc.StackSet))
+	stackUnused := stack.Name != currentStackName && ssc.Traffic != nil && ssc.Traffic[stack.Name].Weight() <= 0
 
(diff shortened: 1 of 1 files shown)


gargravarr:
:+1:

PR #89, opened 2019-01-19
Changed the autoscaler reconciler to use weighted backends

arjunrn:
The `kube-metrics-adapter` now supports weighted backend metrics for RPS scaling. I have change the generation of metrics for the `HorizontalPodAutoscaler` to now generate the weighted metric.


arjunrn:
:+1:

PR #95, opened 2019-02-05
Prescaling and changes to handle errors

arjunrn:
This PR contains the the same changes as the previous one and also contains @muaazsaleem 's fix. I have also made some changes so that the controller does not proceed processing the stackset when one of the step fails. This would prevent traffic from being switched to a stackset which is partially broken. Also added an e2e test for verifying the prescaling logic.


controller/prescale_reconciler.go
```diff
@@ -84,68 +89,41 @@ func (r *PrescaleTrafficReconciler) ReconcileDeployment(stacks map[types.UID]*St
 // minReplicas value is used from the Stack. This means that the HPA is allowed
 // to scale down once the prescaling is done.
 func (r *PrescaleTrafficReconciler) ReconcileHPA(stack *zv1.Stack, hpa *autoscaling.HorizontalPodAutoscaler, deployment *appsv1.Deployment) error {
-	var minReplicas int32
+	hpa.Spec.MaxReplicas = stack.Spec.HorizontalPodAutoscaler.MaxReplicas
```
mikkeloscar: nit: I would just move this to the bottom of the function along with `MinReplicas` maybe `git` just put it here?

controller/stack.go
```diff
@@ -57,11 +57,11 @@ func (c *stacksReconciler) reconcile(ssc StackSetContainer) error {
 		if err != nil {
 			c.recorder.Event(&ssc.StackSet, v1.EventTypeWarning, "GenerateHPA",
 				fmt.Sprintf("Failed to generate HPA %v", err.Error()))
+			return err
```
mikkeloscar: We have to be careful about this case because this change could potentially mean that if one stack is broken nothing else gets reconciled.

There are two conflicting cases (that I can think of) which we should be careful to get right:

1. The newest stack is broken and we want to prevent the controller from moving forward and terminating an old stack.
2. A new stack is broken and the user makes a new deployment with a fix. The broken stack should not prevent the fixed stack from being rolled out (I believe this would happen with this change).

controller/prescale_reconciler.go
```diff
@@ -14,142 +14,78 @@ import (
...
 type PrescaleTrafficReconciler struct {
 	ResetHPAMinReplicasTimeout time.Duration
 }
 
-// ReconcileDeployment prescales the deployment if the prescale annotation is
-// set. Prescale annotation will only be removed from the deployment after it's
-// getting traffic.
+// ReconcileDeployment calculates the number of replicas required when prescaling is active. If there is no associated
+// HPA then the replicas of the deployment are increased directly. Finally once traffic switching is complete the
+// prescaling annotations are removed.
```
mikkeloscar: nit: no longer an annotation.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -203,6 +203,20 @@ type StackSpecTemplate struct {
 	Version string `json:"version"`
 }
 
+// PrescalingStatus contains the prescaling information about the stack
+// +k8s:deepcopy-gen=true
+type PrescalingStatus struct {
+	// LastTrafficIncreaseTime is the timestamp when traffic was last switched to the stack
+	// +optional
+	LastTrafficIncreaseTime string `json:"lastTrafficIncrease,omitempty" protobuf:"bytes,1,opt,name=lastTrafficIncrease"`
```
mikkeloscar: I would use `*metav1.Time` here instead of a string. Then you don't need to convert it in the code. See this as an example: https://github.com/zalando-incubator/stackset-controller/pull/98/files?utf8=%E2%9C%93&diff=unified#diff-79fabbdce31aa152e6981965c59fb575R236

cmd/e2e/prescaling_test.go
```diff
@@ -0,0 +1,226 @@
...
+	require.NoError(t, err)
+
+	// recheck the deployment of the last stack and verify that the number of replicas is the sum of the previous stacks
+	// till the end of the prescaling timeout
+	for i := 1; i <= 6; i++ {
+		deployment, err = waitForDeployment(t, fullThirdStack)
+		require.NoError(t, err)
+		require.EqualValues(t, 6, *(deployment.Spec.Replicas))
+		time.Sleep(time.Second * 10)
+	}
```
mikkeloscar: What's the reason for looping 6 times here? I don't fully understand it since the condition being checked is always `6`.

arjunrn: The prescaling timeout is 60 and the loops tests at 10 second intervals 6 times. I just added it in case the test ever breaks it would be easy to tell why, if it timeout expired or the prescaling didn't happen at all.

controller/prescale_reconciler.go
```diff
@@ -1,155 +1,84 @@
...
 		}
+		stack.Status.Prescaling.Active = true
+		// Update the timestamp in the prescaling information. This bumps the prescaling timeout
+		currentTime := metav1.NewTime(time.Now())
+		stack.Status.Prescaling.LastTrafficIncrease = &currentTime
+	}
 
-		if stack.Spec.HorizontalPodAutoscaler != nil {
-			prescaleReplicas = int32(math.Min(float64(prescaleReplicas), float64(stack.Spec.HorizontalPodAutoscaler.MaxReplicas)))
+	// If prescaling is active and the prescaling timeout has expired then delete the prescaling annotation
```
mikkeloscar: nit: it's not an annotation anymore

mikkeloscar:
:+1:

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #96, opened 2019-02-05
Generate HPA from Stack Autoscaler

arjunrn:
Current the generated HPA is generated directly from the `Autoscaler` defined in the _StackSet` spec. The `Autoscaler` is not copied to the stack. This is means that when the stackset definition is changed all the HPA are changed. This might not be desirable behavior in the case when the pod is resized. In the future the HPA should be an independent resource and should not be changed when the Stackset spec is changed.


mikkeloscar:
lgtm

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #98, opened 2019-02-06
Respect the no-traffic-since annotation when garbage collecting stacks

### LINKED ISSUE #97: ScaledownTTLSeconds should mean "since no traffic" everywhere.
The `ScaledownTTLSeconds` field was intended to have the meaning of scaling down a stack that has not been getting traffic for `ScaledownTTLSeconds`: https://github.com/zalando-incubator/stackset-controller/blob/b528bc912756203470c71bfaa8db777373e8c421/controller/stack.go#L173

However in the stack Garbage collection code it has the meaning of deleting a stack that is not getting traffic and where the CREATION_TIME is older than `ScaledownTTLSeconds`: https://github.com/zalando-incubator/stackset-controller/blob/master/controller/stackset.go#L668

We should ensure that stacks are only deleted if they have not been getting traffic for `ScaledownTTLSeconds` independently of their age.

mikkeloscar:
This makes the garbage collection code respect the `no-time-since`
annotation when considering cleaning up a stack. Before it was using the
stacks creationTime which basically had no meaning for long running
stacks.

Fix #97


gargravarr:
My only question here is: are all those generated headers also intended to be merged? I'm asking because it is not in the PR description. Besides that lgtm.

gargravarr:
👍

mikkeloscar:
> My only question here is: are all those generated headers also intended to be merged? I'm asking because it is not in the PR description. Besides that lgtm.

The generated headers is because of regenerating all the code after updating the stack struct. I could also do the header update in a separate PR, but I guess this is ok as is.

mikkeloscar:
:+1:

mikkeloscar:
@gargravarr I rebased on master (with the e2e test PR merged) so now the generated headers are gone :D

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #101, opened 2019-02-08
Limit the labels used for selectors

mikkeloscar:
This limits the labels used for selectors like the deployment `matchLabel` selectors and service selectors to only the labels which are "owned" by the controller: `stackset` and `stack-version`.

The reason for this change is that we currently inject every label including labels which can change over time like the `deployment-id` (which is used internally in Zalando). Changing the `matchLabels` are not allowed so it makes sense to only use labels which we know will never change.

This will allow redeploying the stackset with the same version as an existing stack in the Zalando environment and will not impact existing stacks as this change only applies to new stacks.


szuecs:
:+1:

szuecs:
Nice set implementation with 0 Byte value :)

mikkeloscar:
:+1:

PR #102, opened 2019-02-25
e2e: add requests/limits to the test container

aermakov-zalando:
`25m`/`128Mi` should probably be enough.


aermakov-zalando:
👍

arjunrn:
:+1:

PR #104, opened 2019-02-26
Fix prescaling when no active stacks present

arjunrn:
There can be case when all the current stacks which are receiving traffic are broken and there are no replicas. In this case the calculated prescaling replicas will be 0 preventing a rollout of the new stack because of HPA validation. I've added a condition to the logic such that the prescaling is performed only when the computed replicas are greater than 0.

### DIFF
--- a/controller/prescale_reconciler.go
+++ b/controller/prescale_reconciler.go
@@ -51,7 +51,7 @@ func (r *PrescaleTrafficReconciler) ReconcileDeployment(stacks map[types.UID]*St
 		}
 
 		// If there is no associated HPA then manually update the replicas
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
LGTM, but would be nice to have a test for it.

aermakov-zalando:
👍

arjunrn:
I could not recreate this condition with tests. I scaled down all active stacks and set the replicas to 0 but the stackset still recovered. So without knowing the exact failure scenario I think this fix is not required. Closing for now.

PR #105, opened 2019-02-27
Fixed the order in which the metrics are sorted

arjunrn:
Because of how the HPA metrics are stored and the how the diff is done for changing the HPA the generated HPAs are continuously updated because the controller thinks it needs to be updated. While this has no downside it's quite bad in terms of calls to the API and updates to the stack object. So I've sorted the metrics in alphabetical order which seems to be how they are stored and retrieved.

### DIFF
--- a/controller/autoscaler.go
+++ b/controller/autoscaler.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"sort"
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #107, opened 2019-03-05
Correctly get ports from pods with multiple containers

mikkeloscar:
This fixes a bug where the stackset controller was unable to create a service for a stack with multiple containers in a pod all exposing an unnamed port.

Before it tried to generate ports for the service of the format: `port-<port-index>` where `port-index` is the index in the port array from the container spec. With two containers both exposing ports it was possible to generate the same port name multiple times which would be rejected by the API validation when submitting the service.

Fix is to change the name format to `port-<container-index>-<port-index>`


Nagogus:
👍

szuecs:
:+1:

mikkeloscar:
:+1:

PR #108, opened 2019-03-06
Fix prescaling when no other stacks receiving traffic

arjunrn:
When the other stacks receiving traffic have no replicas or when there are no stacks currently receiving traffic then the prescaling calculation is incorrect. This PR fixes that issue.

The E2E tests didn't catch this before it happens only in conjunction with `kube-downscaler` and the current stack changing after a deployment.


szuecs:
:+1:

arjunrn:
:+1:

PR #109, opened 2019-03-12
Use resource generation to determine updates

### LINKED ISSUE #49: Use resource generations to determine if resources are up-to-date
> mikkeloscar [22 minutes ago]
> Hey, can I ask you a somewhat related question? Do you know of any functions for injecting default PodTemplateSpec values? I.e. the defaults you get when you submit a pod spec where not everything is defined. Currently I do it like this: https://github.com/zalando-incubator/stackset-controller/blob/2097b1de80710873cdf558bf734dee6e5da848bc/controller/stack.go#L482-L567 but maybe there's a better way?
> 
> liggitt [22 minutes ago]
> for what purpose?
> 
> liggitt [22 minutes ago]
> the defaulting is part of the apiserver and isn't available to external clients
> 
> mikkeloscar [21 minutes ago]
> What I'm trying to achieve is a way to determine if a need to update a resource or not. I have a CRD which "wraps" a deployment and I want to check if the deployment is matching the pod spec defined in the CRD.
> 
> liggitt [14 minutes ago]
> Attempting to match exactly with client side defaulting isn't a reliable way to do that
> 
> liggitt [14 minutes ago]
> You will drift as new fields are added server side
> 
> liggitt [14 minutes ago]
> and not be able to account for changes made by admission plugins, etc
> 
> liggitt [13 minutes ago]
> a more reliable way is to use metadata.generation
> 
> liggitt [12 minutes ago]
> when your custom resource spec changes, it bumps metadata.generation (if you've enabled spec/status for your CRD) (edited)
> 
> liggitt [11 minutes ago]
> your controller can note the observed generation it has reacted to in your custom resource status.observedGeneration
> 
> liggitt [10 minutes ago]
> and record the generation of the resulting deployment it creates in your custom resource as status.deploymentGeneration
> 
> liggitt [9 minutes ago]
> then, whenever your custom resource's metadata.generation != status.observedGeneration or the deployment's metadata.generation != your custom resource's status.deploymentGeneration, your controller can update the deployment spec and update your custom resource's status.deploymentGeneration (edited)
> 
> mikkeloscar [4 minutes ago]
> I didn't consider the admission plugins, that a good point. I will look into the generation idea. Thanks a lot once again!!

mikkeloscar:
Before we tried to determine if a child resource, like a deployment of a
stack, needed updates by comparing the original resource with the
resource after we made potental changes.
This doesn't work in most cases because when a resource is submitted it
potentially goes through admission controllers which changes the
defaults thus modifying the resource. Especially for deployments created
for a stack this didn't work and the stackset-controller is constantly
updating deployment resources even though it's not needed and not actual
changes are applied.

Instead use the Generation on the resources to determine if the child
resource needs an update or not. This is done by storing an annotation
with the "parent" resource generation and then comparing this value when
reconciling a resource.

Fix #49


controller/stack.go
```diff
@@ -353,34 +339,32 @@ func (c *stacksReconciler) manageAutoscaling(stack zv1.Stack, hpa *autoscalingv2
...
-					cmpopts.IgnoreUnexported(resource.Quantity{}),
-				),
-			)
+		stackGeneration := getStackGeneration(hpa.ObjectMeta)
+
+		// only update the resource if there are changes
+		// We determine changes by comparing the stackGeneration
+		// (observed generation) stored on the hpa with the
+		// generation of the Stack.
+		if stackGeneration != stack.Generation {
```
arjunrn: I think the E2E are failing because for prescaling the HPA spec will change with time even when the generation has not changed. So You need to generate the HPA and compare the min and max reeplica fields as well

mikkeloscar: Thanks! I did this extra check for deployment but forgot about the HPA having the same "issue"! :)

mikkeloscar: This is fixed now!

mikkeloscar:
Log reduction in one cluster after changing to this version:

image

Reduction of around 45%

arjunrn:
> Log reduction in one cluster after changing to this version:
> 
> image
> 
> Reduction of around 45%

Somehow I expected the reduction to be much more.

mikkeloscar:
> Somehow I expected the reduction to be much more.

This is in `stups-test` where we seem to have debug logging enabled which is a big contributor I think (didn't look too closely into this yet).

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #111, opened 2019-03-15
Fix dependency on autoscaling package

arjunrn:
Right now `dep check` is complaining that `autoscaling/v1` is used but not managed. So let's just remove it altogether because all our stuff uses `autoscaling/v2` anyway.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #115, opened 2019-04-04
Extracting pure functions from Stack.go/reconcile

muaazsaleem:
At the moment Stack.go/reconcile function both generates the specs for the Deployments/Service & HPA and also applies them. The control flow looks like:
`Stack.go:reconcile` -> `manageDeployments` -> `mangeAutoscaling` ...*

Each step both creates the specs and applies them. 

This design would force us to make changes in many places for the some of the future planned work. To avoid this refactoring the `reconcile` function to extract the following pure function from it:

`func CreateResourcesFromStack(stack StackResource) (DeploymentSpec, ServiceSpec, HPA)`

This would allow us to separate _pure_ logic from applying resources therefore making future changes easier.


controller/stack.go
```diff
@@ -374,67 +374,54 @@ func (c *stacksReconciler) manageService(sc StackContainer, deployment *appsv1.D
...
-		// TODO: "copy" (not move) to newServiceFromStack
-		service.Labels = stack.Labels
-		service.Spec.Selector = limitLabels(stack.Labels, selectorLabels)
-	}
-
-
 	// get service ports to be used for the service
 	var backendPort *intstr.IntOrString
-	if ssc.StackSet.Spec.Ingress != nil {
+	if ssc.StackSet.Spec.Ingress != nil { // TODO: Ask if it's even possible to have a nil Ingress.
```
mikkeloscar: This is technically possible, but not a use case for us. Whether it makes sense to have it or not I'm not sure, but there could be use cases outside of traffic switched applications which we could explore in the future?

controller/stack_resources.go
```diff
@@ -0,0 +1,144 @@
...
+
+func mapCopy(m map[string]string) map[string]string {
+	newMap := map[string]string{}
+	for k, v := range m {
+		newMap[k] = v
+	}
+	return newMap
+}
+
+func newHPAFromStack(stack zv1.Stack) *autoscaling.HorizontalPodAutoscaler {
```
golangcibot: `newHPAFromStack` is unused (from `deadcode`)

controller/stack.go
```diff
@@ -397,81 +374,104 @@ func (c *stacksReconciler) manageService(sc StackContainer, deployment *appsv1.D
...
-				service.Namespace,
-				service.Name,
-			)
-			_, err := c.client.CoreV1().Services(service.Namespace).Update(service)
-			if err != nil {
-				return err
-			}
-		}
+//// manageService manages the service for a given stack.
+func (c *stacksReconciler) manageService2(sc StackContainer, deployment *appsv1.Deployment, ssc StackSetContainer) error {
```
golangcibot: U1000: func `(*stacksReconciler).manageService2` is unused (from `unused`)

muaazsaleem:
👍

controller/stack.go
```diff
@@ -196,7 +172,7 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 			return err
 		}
 	} else {
-		stackGeneration := getStackGeneration(deployment.ObjectMeta)
+		stackGeneration := _getStackGeneration(deployment.ObjectMeta)
```
mikkeloscar: nit: remove the `_` prefix?

controller/stack.go
```diff
@@ -228,6 +204,9 @@ func (c *stacksReconciler) manageDeployment(sc StackContainer, ssc StackSetConta
 		}
 	}
 
+	// TODO: Add this to newDeploymentFromStack once we figure out why it matters that
+	//  it is so late compared to other deployment-related stuff (i.e. why tests fails
+	//  when we put it in this function's first else branch).
```
mikkeloscar: nit: I would remove this TODO as it's a question. Maybe create an issue instead?

controller/stack.go
mikkeloscar: Please remove the `TODO` comment as it's a question.

muaazsaleem: Sounds good will update the comment

controller/stack_resources.go
```diff
@@ -0,0 +1,145 @@
...
+			Selector: limitLabels(stack.Labels, selectorLabels),
+			Type:     v1.ServiceTypeClusterIP,
+			Ports:    servicePorts,
+		},
+	}
+}
+
+// doesStackOwnResource checks whether the stack is assigned to the resource
+// by comparing the stack generation with the corresponding resource annotation.
+func doesStackOwnResource(stack zv1.Stack, resource metav1.Object) bool {
```
mikkeloscar: nit: I think the function name is a bit wrong, it's checking if the stack is up to date or needs updating.

controller/stack_resources.go
```diff
@@ -0,0 +1,145 @@
...
+	decodedGeneration, err := strconv.ParseInt(encodedGeneration, 10, 64)
+	if err != nil {
+		return 0
+	}
+	return decodedGeneration
+}
+
+// assignResourceOwnershipToStack assigns a stack to a resource by specifying the stack's generation
+// in the resource's annotations.
+func assignResourceOwnershipToStack(stack zv1.Stack, resource metav1.Object) {
```
mikkeloscar: nit: again, this is not about ownership but rather if the latest state (generation) of the stack is reflected in the child resource.

controller/stackset.go
```diff
@@ -705,21 +698,23 @@ func (c *StackSetController) getStacksToGC(ssc StackSetContainer) []zv1.Stack {
 
 	// sort candidates by oldest
 	sort.Slice(gcCandidates, func(i, j int) bool {
+		// TODO: maybe we use use noTrafficSince instead of CreationTimeStamp to decide oldest
```
mikkeloscar: nit: `use use` in comment :)

run_e2e.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
mikkeloscar: I would reconsider if we want to include this script as it's very zalando specific. I can see the benefit of this, maybe we can add it to a `script` sub-folder.

muaazsaleem:
Ran go imports already. Don't know how else to make travis happy :(

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #117, opened 2019-04-16
Stacks should be the main entity users interact with

muaazsaleem:
# Stacks should be the main entity

In the current implementation, users create a stackset resource in the cluster which then creates a corresponding stack resource. Deployment of a new version overwrites both the `template` and the `name` in the stackset, causing the controller to create a new stack resource to match. If the stackset is changed without modifying the `name`, the "current" stack will be modified and the previous stacks will be left untouched because the information about their spec is completely lost. This leads to a weird situation where the users have 2 different kinds of stacks, the "current" one that cannot be changed by itself or deleted, and the previous ones which can **only** be changed by modifying the stacks, since the stackset changes no longer affect them. The situation becomes even worse if the users decide to go back to a previous version. An obvious solution would be to just switch traffic to one of the previous stacks and delete the one that's no longer needed, but this doesn't work if it's the "current" one.

Other Kubernetes resources and Senza stacks don't really work like this. In both cases there's one resource that the users interact with. For Kubernetes it's the deployment, which automagically creates and manages the replicasets (and you usually don't fiddle with those). For Senza it's the stacks, and there's no controlling resource for those. We should do something similar for stacksets as well.

@aermakov-zalando  proposes changing the logic to the following one:
 * Stack template in the stackset is only considered when a new stack is created. Further changes to the spec should be ignored or forbidden to avoid confusion.
 * Users are encouraged to think of stacks as the main entities they interact with. They should be able to delete a stack
 * Creation of a stack should be recorded somehow. This will allow deletion of even the "current" stack, since the "current" part is just a small implementation detail of the deployment process.

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -583,9 +583,10 @@ func (c *StackSetController) ReconcileStackSetStatus(ssc StackSetContainer) erro
 	}
 
 	newStatus := zv1.StackSetStatus{
-		Stacks:            int32(len(stacks)),
-		StacksWithTraffic: stacksWithTraffic,
-		ReadyStacks:       readyStacks(stacks),
+		Stacks:               int32(len(stacks)),
+		StacksWithTraffic:    stacksWithTraffic,
+		ReadyStacks:          readyStacks(stacks),
+		ObservedStackVersion: currentStackVersion(stackset),
 	}
 
 	if !equality.Semantic.DeepEqual(newStatus, stackset.Status) {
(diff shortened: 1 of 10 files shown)


controller/stackset.go
```diff
@@ -586,6 +586,9 @@ func (c *StackSetController) ReconcileStackSetStatus(ssc StackSetContainer) erro
 		Stacks:            int32(len(stacks)),
 		StacksWithTraffic: stacksWithTraffic,
 		ReadyStacks:       readyStacks(stacks),
+		// Recording, the creation of stack. This allows deletion of even the "current" stack,
+		//  since the "current" part is just a small implementation detail of the deployment process.
```
mikkeloscar: I think this comment should rather be in the commit/PR description. The problem of "current" stack no longer exists in the code, so it's a bit confusing to have it as a comment IMO.

mikkeloscar:
:+1:

muaazsaleem:
👍

aermakov-zalando:
👍

aermakov-zalando:
LGTM, thanks a lot!

PR #122, opened 2019-04-23
Initial refactor events

njuettner:
Refactoring stackset-controller events

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -152,7 +152,6 @@ func (c *StackSetController) Run(ctx context.Context) {
 			if _, ok := c.stacksetStore[stackset.UID]; ok {
 				if e.Deleted || !c.hasOwnership(&stackset) {
 					delete(c.stacksetStore, stackset.UID)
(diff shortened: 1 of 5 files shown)


controller/stackset.go
```diff
@@ -151,8 +151,8 @@ func (c *StackSetController) Run(ctx context.Context) {
 			// update/delete existing entry
 			if _, ok := c.stacksetStore[stackset.UID]; ok {
 				if e.Deleted || !c.hasOwnership(&stackset) {
-					c.recorder.Eventf(e.StackSet, apiv1.EventTypeNormal, "DeleteStackSet", "StackSet '%s/%s' deleted, removing references", stackset.Namespace, stackset.Name)
 					delete(c.stacksetStore, stackset.UID)
+					c.recorder.Eventf(e.StackSet, apiv1.EventTypeNormal, "DeletedStackSet", "StackSet '%s/%s' deleted, removing references", stackset.Namespace, stackset.Name)
```
mikkeloscar: I think it makes sense to drop this event because it will trigger when a stackset is not owned by the stackset-controller because of the `hasOwnership` check, or when a stackset is deleted in which case it probably doesn't make sense to emit an event for it.

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #125, opened 2019-04-26
Only applying StackLifeCycle.Limit to stacks without traffic and enforcing it to >= 1

muaazsaleem:
Changing the semantics of StackLifeCycle.Limit to exclude the stacks
that are getting traffic. So `StackLifeCycle.Limit: 1` would now mean, keep
at least 2 stacks. This is more intuitive because the stacks that are getting
traffic would never be deleted anyways.

Also modifying the CRD definition to enforce StackLifeCycle.Limit >= 1.
This is so people have at least 1 stack around to safely switch to. Since
HPAs are now not deleted, this extra stack should get safely downscaled
without costing too much.

Signed-off-by: Muhammad Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/cmd/e2e/ttl_test.go
+++ b/cmd/e2e/ttl_test.go
@@ -91,6 +91,7 @@ func TestStackTTLWithIngress(t *testing.T) {
 	}
 }
 
+// TestStackTTLForLatestStack tests that the latest stack gets scaled down and isn't treated differently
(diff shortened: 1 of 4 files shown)


muaazsaleem:
👍

mikkeloscar:
:+1:

PR #128, opened 2019-05-08
Not switching traffic until all the prescaling stacks are ready

arjunrn:
When there are multiple stacks which are prescaled and one of them doesn't have all it's replicas ready then traffic is only sent to the one with all replicas ready. This is an issue because both could have been receiving traffic and suddenly only one will get all the traffic.

### DIFF
--- a/controller/prescale_reconciler.go
+++ b/controller/prescale_reconciler.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"fmt"
 	"math"
 	"time"
 
@@ -19,6 +20,14 @@ const (
 
 type PrescaleTrafficReconciler struct {
 	ResetHPAMinReplicasTimeout time.Duration
+}
+
+type trafficSwitchingError struct {
+	err string
+}
+
+func (t trafficSwitchingError) Error() string {
+	return t.err
 }
 
 // ReconcileDeployment calculates the number of replicas required when prescaling is active. If there is no associated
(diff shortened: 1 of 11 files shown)


controller/prescale_reconciler_test.go
```diff
@@ -1174,12 +1174,102 @@ func TestReconcileIngressTraffic(tt *testing.T) {
...
 		tt.Run(ti.msg, func(t *testing.T) {
 			trafficReconciler := PrescaleTrafficReconciler{}
-			availableWeights, allWeights := trafficReconciler.ReconcileIngress(ti.stacks, ti.ingress, ti.traffic)
+			availableWeights, allWeights := trafficReconciler.ReconcileIngress(ti.stacks, ti.traffic)
 			require.Equal(t, ti.expectedAvailableWeights, availableWeights)
 			require.Equal(t, ti.expectedAllWeights, allWeights)
 		})
 	}
 }
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

muaazsaleem:
We still need to show the user a warning if traffic switching never happens because a stack is not ready.

muaazsaleem:
Added an e2e test for the scenario, it seems that the code has a bug. Trying to reproduce it with a unit-test to make progress.

controller/prescale_reconciler.go
```diff
@@ -108,18 +118,25 @@ func (r *PrescaleTrafficReconciler) ReconcileIngress(stacks map[types.UID]*entit
 
 		// prescale if stack is currently less than desired traffic
 		if traffic[stack.Stack.Name].ActualWeight < traffic[stack.Stack.Name].DesiredWeight && deployment != nil {
+			var actualReplicas int32 = 1
+			var desiredReplicas int32 = 1
```
golangcibot: ineffectual assignment to `desiredReplicas` (from `ineffassign`)

muaazsaleem:
👍

controller/prescale_reconciler_test.go
```diff
@@ -1167,20 +1170,305 @@ func TestReconcileIngressTraffic(tt *testing.T) {
...
 	} {
 		tt.Run(ti.msg, func(t *testing.T) {
 			trafficReconciler := PrescaleTrafficReconciler{}
-			availableWeights, allWeights := trafficReconciler.ReconcileIngress(ti.stacks, ti.ingress, ti.traffic)
+			availableWeights, allWeights, err := trafficReconciler.ReconcileIngress(ti.stacks, ti.ingress, ti.traffic)
 			require.Equal(t, ti.expectedAvailableWeights, availableWeights)
 			require.Equal(t, ti.expectedAllWeights, allWeights)
+
+			// Check err later because the weights returned still matter
+			if ti.err != nil {
```
szuecs: you can compare `err` with `ti.err`

szuecs:
:+1:

controller/prescale_reconciler_test.go
muaazsaleem: Noted, will fix in a follow up PR

aermakov-zalando:
👍

PR #129, opened 2019-05-08
Correctly inject stack labels to pod template

mikkeloscar:
Correctly inject labels to the pod template instead of overwriting the labels on deployment creation. If we don't do this creating a deployment will result in any user defined labels being dropped. This is fixed as soon as the deployment is updated but results in creating a new ReplicaSet which is confusing and triggers a rolling update of the deployment.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #130, opened 2019-05-09
Divide controller überpackage into smaller semantic packages

thilp:
Before:
```
	controller [depends on itself for everything]
```
After:
```
	controller	[depends on entities]
	    |
	 entities	[no dependencies]
```

This allows us to extend the TrafficReconciler interface in another package (like `canary`) without creating circular dependencies (`controller` initializes `canary.SomeType`, `canary` uses `controller.SomeContainer`).


thilp:
The added lines in `stack.go` used to belong (mostly) to `stack_resources.go` before the latter moved in the new `entities` package.

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #135, opened 2019-05-28
Reset the prescaling replica count when prescaling multiple times

arjunrn:
There was a bug in the prescaling logic where the previously calculated replicas was not reset and subsequent prescaling calculations would be added to past calculations.


mikkeloscar:
:+1:

muaazsaleem:
👍

PR #136, opened 2019-05-28
Return error right away to prevent side effects

mikkeloscar:
Moves the error handling to where the error is returned. This ensures that we don't modify the `ingress` variable even when an error is returned.

In the actual code this didn't matter because it would be handled later on. But it makes it simpler to follow the code IMO and allows the `ReconcileIngress` implementation to return `nil` for the traffic maps in case of an error without crashing the program.


mikkeloscar:
:+1:

lispyclouds:
👍

muaazsaleem:
The problem was that `reconcileIngress` would still try to create/apply the `Ingress`  after the err: https://github.com/zalando-incubator/stackset-controller/blob/master/controller/ingress.go#L121
If we return right away the Ingress won't be fully ready.

mikkeloscar:
Closing in favor of #137

PR #137, opened 2019-05-29
Refactor the core logic to split updating the resources and reconciliation logic

aermakov-zalando:
This is a huge refactoring of the code base that tries to split the part that syncs with Kubernetes from the part that updates the state of the stacksets, stacks and other related objects, and hopefully tries to make the second part smaller and easier to understand and test.


pkg/core/types.go
```diff
@@ -0,0 +1,228 @@
...
+		if weights, ok := ssc.Ingress.Annotations[backendWeightsAnnotationKey]; ok {
+			err := json.Unmarshal([]byte(weights), &actual)
+			if err != nil {
+				return fmt.Errorf("failed to get current actual Stack traffic weights: %v", err)
+			}
+		}
+
+		// Remove weights for stacks that no longer exist, normalize the result
+		for _, weights := range []map[string]float64{desired, actual} {
+			for name, _ := range weights {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for name := range ...` (from `golint`)

pkg/core/stack_resources.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+	}
+	return template
+}
+
+func (sc *StackContainer) resourceMeta() metav1.ObjectMeta {
+	resourceLabels := mapCopy(sc.Stack.Labels)
+
+	return metav1.ObjectMeta{
+		Name:        sc.Stack.Name,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/core/stackset.go
```diff
@@ -0,0 +1,266 @@
+package core
+
+import (
+	"encoding/json"
+	"errors"
+	"sort"
+
+	zv1 "github.com/zalando-incubator/stackset-controller/pkg/apis/zalando.org/v1"
+	"k8s.io/api/core/v1"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

controller/stack_resources.go
```diff
@@ -0,0 +1,221 @@
+package controller
+
+import (
+	"github.com/zalando-incubator/stackset-controller/pkg/core"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/core/stack_resources_test.go
```diff
@@ -138,30 +122,27 @@ func TestGetServicePorts(tt *testing.T) {
...
-					PodTemplate: v1.PodTemplateSpec{
-						Spec: v1.PodSpec{
-							Containers: []v1.Container{
-								{
-									Ports: []v1.ContainerPort{
-										{
-											Name:          "ingress-invalid",
-											ContainerPort: 8080,
-										},
+			msg:
```
golangcibot: File is not `goimports`-ed (from `goimports`)

controller/stackset.go
```diff
@@ -504,248 +444,281 @@ func (c *StackSetController) del(obj interface{}) {
...
+			container.StackSet,
+			v1.EventTypeWarning,
+			"FailedSwitchTraffic",
+			err.Error())
 	}
 
-	return nil
-}
+	// Mark stacks that should be removed
+	err = container.MarkForRemoval()
```
muaazsaleem: Can we call this method "MarkExpiredStackForRemoval" or sth like that cuz it reads like we're marking the `container` itself for removal (the comment does help btw)

muaazsaleem: maybe `MarkRemovableStacks`

controller/stackset.go
```diff
@@ -504,248 +444,281 @@ func (c *StackSetController) del(obj interface{}) {
...
-	stackset := ssc.StackSet
-	c.recorder.Eventf(&stackset,
-		apiv1.EventTypeWarning,
-		"FailedUpdateStack",
-		"Failed to create/update stack: %v", err)
-}
+// CreateCurrentStack creates a new Stack object for the current stack, if needed
+func (c *StackSetController) CreateCurrentStack(ssc core.StackSetContainer) error {
+	newStack, newStackVersion := ssc.NewStack()
+	if newStack == nil {
```
muaazsaleem: I assume `nil` here means, a stack creation or update isn't needed?

muaazsaleem: I see a `nil` actually means a _stack creation_ isn't needed

pkg/core/stackset.go
```diff
@@ -0,0 +1,266 @@
...
+func (ssc *StackSetContainer) NewStack() (*StackContainer, string) {
+	stackset := ssc.StackSet
+
+	observedStackVersion := stackset.Status.ObservedStackVersion
+	stackVersion := currentStackVersion(stackset)
+	stackName := generateStackName(stackset, stackVersion)
+
+	stack := ssc.stackByName(stackName)
+
+	if stack == nil && observedStackVersion != stackVersion {
```
muaazsaleem: Should we document that "A stack creation is needed only in case if it was never created or the spec has been updated requiring a new stack". I just had to recall that that's why we're checking both `stack` and `observedStackVersion`

controller/stack_resources.go
```diff
@@ -0,0 +1,221 @@
...
+		service.Name)
+	return nil
+}
+
+func (c *StackSetController) ReconcileStackIngress(sc *core.StackContainer) error {
+	ingress, err := sc.GenerateIngress()
+	if err != nil {
+		return err
+	}
+	existing := sc.Resources.Ingress
```
muaazsaleem: We never really set this field to `nil` should we? We also don't have a "marked for deletion" field like we do for `stacks`

aermakov-zalando: `StackContainers` are repopulated from scratch on every controller loop iteration, so we don't really need to keep `Resources` in sync.

aermakov-zalando: I'll add some comments about it, but please note that the rest of the code doesn't actually update anything!

pkg/core/traffic.go
```diff
@@ -0,0 +1,163 @@
...
+		if stack.HasTraffic() {
+			stack.noTrafficSince = time.Time{}
+		} else if stack.noTrafficSince.IsZero() {
+			stack.noTrafficSince = time.Now()
+		}
+	}
+	return err
+}
+
+// fallbackStack returns a stack that should be the target of traffic if none of the existing stacks get anything
```
muaazsaleem: Is this a change in behaviour or was this happening before too? If this is new then we should try to avoid this situation from ever happening instead of picking a stack once it's already happened. See https://docs.google.com/document/d/1IIXH6_EeL-e785RdfI5Pqx1FFZcFPLHNBUCaoL4px_M/edit#heading=h.apqxhe8vdunt

pkg/core/traffic_prescaling.go
```diff
@@ -0,0 +1,98 @@
...
+	// * If no stacks are currently being prescaled fall back to the current weights.
+	// * If no stacks are getting traffic fall back to desired weight without checking health.
+	currentWeights := make(map[string]float64, len(stacks))
+	actualWeights := make(map[string]float64, len(stacks))
+	for _, stack := range stacks {
+		currentWeights[stack.Stack.Name] = stack.actualTrafficWeight
+
+		deployment := stack.Resources.Deployment
+
+		if stack.actualTrafficWeight < stack.desiredTrafficWeight && stack.deploymentUpdated {
```
muaazsaleem: Are we using `stack.deploymentUpdated` because it won't be `nil` even if the user deleted the `deployment`?

pkg/core/traffic_prescaling.go
```diff
@@ -0,0 +1,98 @@
...
+			if stack.Stack.Status.Prescaling.Active {
+				var desired int32 = 1
+				if deployment.Spec.Replicas != nil {
+					desired = *deployment.Spec.Replicas
+				}
+
+				if desired >= stack.Stack.Status.Prescaling.Replicas && deployment.Status.ReadyReplicas >= stack.Stack.Status.Prescaling.Replicas {
```
muaazsaleem: In my understanding we're missing the case handled here*: https://github.com/zalando-incubator/stackset-controller/pull/128/files#diff-45c494197cc2222445770667294f724fR153
because when the weights get normalized, the stacks that are not ready yet will get their traffic overwritten to 0.

* If stack is getting traffic but ReadyReplicas < prescaleReplicas, don't remove traffic from it.

muaazsaleem: I think have the following `else` statement will fix that
```
else if desired >= stack.Stack.Status.Prescaling.Replicas && deployment.Status.ReadyReplicas *<* stack.Stack.Status.Prescaling.Replicas {
	actualWeights[stack.Stack.Name] = stack.actualTrafficWeight
}
```

pkg/core/types.go
```diff
@@ -0,0 +1,228 @@
...
+	return sc.actualTrafficWeight > 0 || sc.desiredTrafficWeight > 0
+}
+
+func (sc *StackContainer) IsReady() bool {
+	// Haven't updated yet
+	if !sc.deploymentUpdated {
+		return false
+	}
+
+	return sc.deploymentReplicas == sc.updatedReplicas && sc.deploymentReplicas == sc.readyReplicas
```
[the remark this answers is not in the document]
aermakov-zalando: Because ready doesn't imply updated. These could be pods using the previous configuration, and if you rescale the deployment at the same time, we don't want to proceed.

controller/stackset.go
aermakov-zalando: Yes, we don't update stacks anymore.

pkg/core/stackset.go
aermakov-zalando: Added a comment.

controller/stackset.go
aermakov-zalando: Done.

pkg/core/traffic.go
aermakov-zalando: Previously it would do different (and completely unspecified) things depending on the traffic reconciler. Unfortunately, we can't really avoid this situation completely, and we still have to do something to ensure that 100% of traffic goes somewhere. This happens, for example, when the stackset is originally created (initial stack should get traffic), or if the user deletes the only stack that currently gets traffic.

pkg/core/traffic_prescaling.go
aermakov-zalando: Existing code was testing for `deployment != nil`, I haven't really touched that part yet.

pkg/core/traffic_prescaling.go
aermakov-zalando: I haven't picked up the recent changes yet, will do soon.

pkg/core/traffic.go
muaazsaleem: Do you think it would be too much work to add a special condition for when the stackset is originally created and for the other case, just not allow users to delete a stack that's getting traffic?

aermakov-zalando: We cannot really do that (or maybe we could, but it'd be a lot more complicated).

pkg/core/traffic_test.go
```diff
@@ -0,0 +1,667 @@
...
+	zv1 "github.com/zalando-incubator/stackset-controller/pkg/apis/zalando.org/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var (
+	hourAgo        = time.Now().Add(-time.Hour)
+	fiveMinutesAgo = time.Now().Add(-5 * time.Minute)
+)
+
+func TestTrafficSwitchNoIngress(t *testing.T) {
```
arjunrn: Which scenario is this test for? Is this for when an ingress is accidentally deleted?

pkg/core/traffic_prescaling.go
```diff
@@ -0,0 +1,93 @@
+package core
+
+import (
+	"sort"
+	"strings"
+	"time"
+)
+
+// PrescalingTrafficReconciler is the most simple traffic reconciler which
```
arjunrn: Is it though :)

pkg/core/traffic_prescaling.go
```diff
@@ -0,0 +1,93 @@
...
+	var nonReadyStacks []string
+	actualWeights := make(map[string]float64, len(stacks))
+	for stackName, stack := range stacks {
+		// Check if we're increasing traffic but the stack is not ready
+		if stack.desiredTrafficWeight > stack.actualTrafficWeight {
+			var desiredReplicas = stack.deploymentReplicas
+			if stack.prescalingActive {
+				desiredReplicas = stack.prescalingReplicas
+			}
+			if !stack.IsReady() || stack.updatedReplicas < desiredReplicas || stack.readyReplicas < desiredReplicas {
```
arjunrn: Because of the condition `stack.updatedReplicas < desiredReplicas` if a stack is updated in place then won't be suddenly be marked as _nonReady_ even if is getting a significant portion of the traffic?

arjunrn: nvm, I see that it will return an error later

pkg/core/traffic_test.go
aermakov-zalando: It's to test that everything is correctly cleaned up if the user disables ingresses completely.

pkg/core/traffic_prescaling.go
aermakov-zalando: It would, but it only means that we won't be switching traffic in this iteration. All existing weights would stay in place.

pkg/core/traffic_prescaling.go
aermakov-zalando: Fixed!

pkg/core/traffic_prescaling.go
aermakov-zalando: This should be fixed now that we don't switch traffic until the stacks are ready anymore.

pkg/core/traffic_prescaling.go
```diff
@@ -0,0 +1,98 @@
...
+	// Prescale stacks if needed
+	for _, stack := range stacks {
+		// If traffic needs to be increased
+		if stack.desiredTrafficWeight > 0 && stack.desiredTrafficWeight > stack.actualTrafficWeight {
+			// If prescaling is not active then calculate the replicas required
+			if !stack.prescalingActive {
+				stack.prescalingReplicas = totalReplicas
+
+				// If no other stacks are currently active
+				if stack.prescalingReplicas == 0 {
```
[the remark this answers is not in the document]
aermakov-zalando: Because we're assigning it from `totalReplicas` in the previous line. This is our fallback for when we can't calculate how much to prescale for, so we just fallback to `stack.Spec.Replicas`.

controller/stackset.go
```diff
@@ -254,114 +198,105 @@ func (c *StackSetController) collectResources() (map[types.UID]*entities.StackSe
...
+Items:
 	for _, stack := range stacks.Items {
 		if uid, ok := getOwnerUID(stack.ObjectMeta); ok {
 			if s, ok := stacksets[uid]; ok {
-				s.StackContainers[stack.UID] = &entities.StackContainer{Stack: stack}
+				stack := stack
+				s.StackContainers[stack.UID] = &core.StackContainer{
+					Stack: &stack,
+				}
+				continue Items
```
mikkeloscar: Isn't `continue` enough here? No need to define `Items:`

aermakov-zalando: Fixed.

mikkeloscar:
Awesome! Makes the code much simpler to follow! It's a bit hard to review such a big PR in details but I like the overall structure and trust that the e2e tests validate that not logic was broken. :P

aermakov-zalando:
@mikkeloscar Yeah, I know, sorry. :( I couldn't figure out how to do it incrementally, so I decided to just do it in one painful go.

mikkeloscar:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #138, opened 2019-06-20
Prescaling: calculate replicas based on existing stacks

aermakov-zalando:
Calculate the number of replicas required when prescaling more intelligently. Instead of just adding up all existing pods and using it as the target number, no matter the desired traffic weight, try to derive how many replicas we need based on the current stacks and the target traffic weight. For example, if we find out that 10% of traffic can be handled by 1 pod and the desired weight is 30%, we only prescale up to 3 replicas.

**TBD**: rebase on `master` after the main refactoring PR is merged.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #144, opened 2019-07-09
Improve error handling

aermakov-zalando:
Instead of just aborting on the first encountered error, try to proceed and reconcile the remaining resources. The only thing that currently causes a hard error is `UpdateFromResources()`, the rest should be logged, reported as an event and ignored.

`StackContainer` logic has been adjusted to check other stack resources in `IsReady()` as well to prevent a situation where traffic is directed to a stack that can't be reconciled.

### DIFF
--- a/cmd/e2e/broken_stack_test.go
+++ b/cmd/e2e/broken_stack_test.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func TestBrokenStacks(t *testing.T) {
+	t.Parallel()
+
+	stacksetName := "stackset-broken-stacks"
+	factory := NewTestStacksetSpecFactory(stacksetName).Ingress().StackGC(1, 30)
+
+	firstVersion := "v1"
+	firstStack := fmt.Sprintf("%s-%s", stacksetName, firstVersion)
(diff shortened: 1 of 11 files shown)


muaazsaleem:
> The only thing that currently causes a hard error is UpdateFromResources()

Could you say why `UpdateFromResources()` and `ReconcileStatuses()` cause hard errors?

muaazsaleem:
It's clear why the `container` parameter in  `ReconcileStackSet(container *core.StackSetContainer)` is a pointer (cuz of `container.MarkExpiredStacks()`), why were the parameters of other methods changed to pointers? Maybe there is a golang best practice I don't know about?

aermakov-zalando:
@muaazsaleem `ReconcileStatuses()` is the last call, so there's no point in trying to do anything. It also shouldn't be possible to break it permanently by a user action.

We can't really proceed if `UpdateFromResources()` fails because the whole state will be completely broken. Note that currently it only fails if it can't parse traffic information, which can only happen if the users deliberately corrupt the annotations. We could implement a fallback strategy for this, but I'm not sure if it's worth it.

muaazsaleem:
> It's clear why the container parameter in ReconcileStackSet(container *core.StackSetContainer) is a pointer (cuz of container.MarkExpiredStacks()), why were the parameters of other methods changed to pointers? Maybe there is a golang best practice I don't know about?

I think my question is too generic, I'll follow up with more specific ones if I have them

muaazsaleem:
👍

muaazsaleem:
👍

muaazsaleem:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #148, opened 2019-08-15
Don't set replicas to nil when updating a deployment

aermakov-zalando:
Turns out that setting a field to `nil` when doing an `Update()` doesn't mean that the current value should be preserved. Instead, the API server runs the default logic for it, so every time we update a deployment we also reset replicas to `1`. Fix by keeping the current amount of replicas instead.

Supersedes #147.


aermakov-zalando:
👍

arjunrn:
👍

PR #157, opened 2019-08-21
Don't consider scaled down stacks ready

aermakov-zalando:
Stacks with 0 replicas are treated as ready, so we switch traffic immediately, which is obviously a problem.


arjunrn:
👍

aermakov-zalando:
👍

PR #171, opened 2019-10-24
Use averageValue for skipper based autoscaler

mikkeloscar:
This sets the field `averageValue` for the skipper based autoscaler.

This requires
https://github.com/zalando-incubator/kube-metrics-adapter/releases/tag/v0.0.4
which adds support for `averageValue` such that average is calculated by
the HPA rather than by the kube-metrics-adapter. This makes the
resulting average more reliable for scaling.

Note this means the stackset-controller only supports Kubernetes
`>=v1.14` if the skipper autoscaler is used. (Versions older than `v1.14` are not officially supported anymore anyway).


mikkeloscar:
:+1:

szuecs:
:+1:

PR #174, opened 2019-11-06
Enable scale command to be used for stacks

mikkeloscar:
This allows scaling stacks via:

```
kubectl scale stack <stackName> --replicas=n
```

This would not work for HPA enabled stacks so only useful if you don't have an HPA.

There is mentioned support for HPA in the docs: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource via the `LabelSelectorPath` value. But I don't understand yet how this works.

### DIFF
--- a/docs/stack_crd.yaml
+++ b/docs/stack_crd.yaml
@@ -49,6 +49,7 @@ spec:
     scale:
       specReplicasPath: .spec.replicas
       statusReplicasPath: .status.replicas
+      labelSelectorPath: .status.labelSelector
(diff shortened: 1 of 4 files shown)


szuecs:
@mikkeloscar can you please add a test for the subresource?

mikkeloscar:
Expanded the stack status test to include the `LabelSelector`.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #175, opened 2019-11-12
Don't send stack status updates unless necessary

aermakov-zalando:
This was done for stackset statuses, but stack statuses were always updated.


aermakov-zalando:
👍

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #179, opened 2019-12-12
Add Update Strategy to the Stack

arjunrn:
This PR adds `UpdateStrategy` to the `StackSpec` which means an update strategy can be specified as part of the stack template which will be propagated to the deployment.

### DIFF
--- a/pkg/core/stack_resources.go
+++ b/pkg/core/stack_resources.go
@@ -166,7 +166,12 @@ func (sc *StackContainer) GenerateDeployment() *appsv1.Deployment {
 		updatedReplicas = wrapReplicas(sc.deploymentReplicas)
 	}
 
-	return &appsv1.Deployment{
+	var strategy *appsv1.DeploymentStrategy
+	if stack.Spec.Strategy != nil {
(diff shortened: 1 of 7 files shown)


pkg/core/stack_resources.go
```diff
@@ -166,7 +167,13 @@ func (sc *StackContainer) GenerateDeployment() *appsv1.Deployment {
 		updatedReplicas = wrapReplicas(sc.deploymentReplicas)
 	}
 
-	return &appsv1.Deployment{
+	var strategy *appsv1.DeploymentStrategy
+	log.Infof("Stack %s/%s update strategy: %#v", stack.Namespace, stack.Name, stack.Spec.Strategy)
```
mikkeloscar: This looks like a debug log. Maybe we can remove it considering we also don't log other sort of values in this way?

arjunrn: Yes it's a debug log :man_facepalming:

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #180, opened 2019-12-16
feature: enable other controller ingress annotations, fixes #113

### LINKED ISSUE #113: Add support for other ingress controller
Current the controller is tied to _Skipper_ as an ingress controller and the dependency is for the traffic switching annotation. The controller should support other ingress providers as well such like _Traefik_ which works exactly the same as skipper with the annotation `traefik.ingress.kubernetes.io/service-weights`. _Nginx_ behaves a little differently because it doesn't define the multiple backends on the same ingress but allows multiple ingresses with the same host.

szuecs:
feature: enable other controller ingress annotations, fixes #113


mikkeloscar:
:+1:

szuecs:
:+1:

PR #182, opened 2019-12-17
Generate CRDs

arjunrn:
Generate CRDs using tools

Updated:
* Set `preserveUnknownFields: false` such that unknown fields gives a validation error.
* Remove `x-kubernetes-list-map-keys:` which doesn't work: https://github.com/kubernetes-sigs/controller-tools/issues/444


pkg/apis/zalando.org/v1/types.go
```diff
@@ -179,19 +187,25 @@ type ActualTraffic struct {
...
 
 // DesiredTraffic is the desired traffic setting to direct traffic to
 // a stack. This is meant to use by clients to orchestrate traffic
 // switching.
 type DesiredTraffic struct {
-	StackName string  `json:"stackName"`
-	Weight    float64 `json:"weight"`
+	StackName string `json:"stackName"`
+	// +kubebuilder:validation:Type=number
+	// +kubebuilder:validation:Format=float
```
muaazsaleem, line 202: nit: Is this deliberately not `float64`? i.e // +kubebuilder:validation:Format=float64

muaazsaleem: Ah nvm, I see [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaObject) that it's not a valid format.

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #183, opened 2019-12-19
Fixed CRD issues and added update strategy

arjunrn:
The CRD had some incorrectly named and missing field. They have been added. Also added the `strategy` field which was introduced in #179


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #184, opened 2019-12-20
More fixes for CRDs

arjunrn:
All objects with a `properties` field have been given an `type: object` field to pass schema validation.
Also all types of `IntStr` need the field `x-kubernetes-int-or-str` and the types specified in that order.


mikkeloscar:
I trust you :D

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #186, opened 2020-01-07
Event recorder: increase the QPS

aermakov-zalando:
Default configuration for event recorder QPS is 1 event in 5 minutes, with 25 burst credit. This means that if there's an ongoing condition (e.g. a problem with an ingress, we'll send an event every 10 seconds in the first ~4 minutes, but then suddenly drop down to only updating the event once every 5 minutes), which means they'll be ignored by CDP.

We should think about switching to something like Conditions in the future for reporting whether the child resources are fine, but increasing the QPS seems like a cheap and easy interim fix.


aermakov-zalando:
👍

njuettner:
👍

PR #187, opened 2020-01-08
Suppress meaningless logs from client-go

mikkeloscar:
`client-go` produces logs like below when a watch times out. These logs just cause noise, so there is no need to see them. We can suppress them via this hack where we replace `klog` with an empty implementation.

Example log:

```
W0108 08:55:20.748693       1 reflector.go:340] pkg/mod/k8s.io/client-go@v0.0.0-20191202165906-e3970fe37d8a/tools/cache/reflector.go:108: watch of *v1.StackSet ended with: an error on the server ("unable to decode an event from the watch stream: net/http: request canceled (Client.Timeout exceeded while reading body)") has prevented the request from succeeding
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,3 +29,5 @@ require (
 	k8s.io/code-generator v0.0.0-20191004115455-8e001e5d1894
 	k8s.io/gengo v0.0.0-20191120174120-e74f70b9b27e // indirect
 )
+
+replace k8s.io/klog => github.com/mikkeloscar/knolog v0.0.0-20190326191552-80742771eb6b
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Can we hide only these messages instead of just disabling all logging for client-go? There might be useful stuff in there, even if it's not there now it they could add it in the future.

mikkeloscar:
It's messages like this: https://github.com/kubernetes/client-go/blob/8ead54f5cff8df71cc709c009970518a2beec9cc/tools/cache/reflector.go#L340 so not simple to filter out unless we have an implementation which does string matching or so. I don't really see the need for any of those logs even possible future logs as I would expect an error to be returned if we really have to do something. So I would rather leave it as is or block all instead of trying to track which specific ones we want to suppress.

aermakov-zalando:
`klog` can have different logging levels for files, or does that not work for some reason?

aermakov-zalando:
I.e. we could just set logging level for `reflector` to error only.

mikkeloscar:
> klog can have different logging levels for files, or does that not work for some reason?

Sorry, I'm not aware of this feature and can't seem to find any docs or example. Do you know how it works?

aermakov-zalando:
@mikkeloscar [klog.go](https://github.com/kubernetes/klog/blob/master/klog.go#L64). I'm not sure how well flag would work with kingpin, though, but we could probably just configure it in our `main()`.

mikkeloscar:
I will try that

mikkeloscar:
`klog` only exposes the setting via flags: https://github.com/kubernetes/klog/blob/c4f748769d6e4fc2c4ef1b720dbd1654e427ea48/klog.go#L417-L435 so I don't see any way to actually do it when we use kingpin.

aermakov-zalando:
We could try changing `os.Args` after `kingpin.Parse()` and then run `flag.Parse()`, for example.

PR #189, opened 2020-01-13
Actually remove the annotation during migration

aermakov-zalando:
Setting it to an empty string doesn't actually remove it.


muaazsaleem:
👍

aermakov-zalando:
👍

aermakov-zalando:
👍

PR #190, opened 2020-01-13
Migration: don't crash on stacksets without ingresses

aermakov-zalando:
We do have stacksets without an associated ingress sometimes.


mikkeloscar:
:+1:

muaazsaleem:
👍

aermakov-zalando:
👍

PR #191, opened 2020-01-15
fix migrations to stackset

szuecs:
fix: `--migrate-to=stackset` crashed if there was not ownerReference set 
fix: have only one place for the const definition, such that overrides also work as expected


szuecs:
:+1:

PR #192, opened 2020-01-21
Statuses: drop omitempty where it doesn't make sense

aermakov-zalando:
Keeping explicit zeroes in the status objects makes it easier to read the statuses, and it's impossible to print these values nicely if they're unset. Drop `omitempty` from the places where it doesn't make sense.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #193, opened 2020-01-21
Populate stack.status.desiredReplicas from the deployment

aermakov-zalando:
Apparently this number was populated only from the HPA, and was left at 0 if HPA was unused. It's unclear whether something like this is useful to anyone, so let's populate it from `deployment.spec` instead. It'll show exactly the same, except for cases when the HPA controller fails, and all the cases where it's currently broken (e.g. no HPA).


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #194, opened 2020-01-24
Drop support for traffic management via Ingresses

aermakov-zalando:
We've migrated to stackset-based traffic management and don't intend to migrate back. Let's just drop the unnecessary code.


muaazsaleem:
👍

mikkeloscar:
FYI: I changed the version bump from `v2.0` to `v1.3` i.e. just bumping the minor version. I think this is the right way because while I understand that we change the API a bit, we're not doing major changes overall to StackSet which would prompt a v2 in my mind.

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #197, opened 2020-02-05
Make sure ttl is handled correctly for stacksets with externalIngress

mikkeloscar:
In case a stackset uses `externalIngress` the stack Lifecycle is not handled correctly when updating the stackset to a new version.

This happens because the Garbage collection logic just looks if the stack has an ingress which is not the case for `externalIngress` stacksets. We need to check both for ingress AND externalIngress when deciding on stack cleanup.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #202, opened 2020-02-25
Stack ingresses: use an explicit cluster domain

aermakov-zalando:
Use an explicitly provided cluster domain when generating per-stack ingresses (so just `<stack-name>.<cluster-domain>` instead of going through all of the main ingress hostnames and generating `<stack-name>.<domain>`). The latter either produced identical results, duplicate entries or didn't work anyway.


mikkeloscar:
:+1:

aermakov-zalando:
:-1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #203, opened 2020-02-25
Report errors via errors counter metrics

linki:
This adds an ErrorHandler that increases a Prometheus metric `stackset_errors_count` when a Kubernetes-internal error occurs.

This type of error handling is used by some Kubernetes components, e.g. some parts of `client-go`, to log errors that would otherwise be ignored because they cannot be surfaced to a user, e.g. in a control loop:

```bash
E0225 17:43:56.916702   12298 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191202165906-e3970fe37d8a/tools/cache/reflector.go:108: Failed to list *v1.StackSet: v1.StackSetList.Items: []v1.StackSet: v1.StackSet.Spec: v1.StackSetSpec.StackTemplate: v1.StackTemplate.Spec: v1.StackSpecTemplate.StackSpec: PodTemplate: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Resources: v1.ResourceRequirements.Requests: Limits: unmarshalerDecoder: unable to parse quantity's suffix, error found in #10 byte of ...|y":"320mi"},"request|..., bigger context ...|esources":{"limits":{"cpu":"500m","memory":"320mi"},"requests":{"cpu":"500m","memory":"320Mi"}}}]}},|...
```

The error above would lead to the counter increasing.

Note that not all Kubernetes components use this kind of error handling so it doesn't catch all errors.

### DIFF
--- a/pkg/core/metrics.go
+++ b/pkg/core/metrics.go
@@ -4,6 +4,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 )
 
 const (
@@ -11,6 +12,7 @@ const (
 
 	metricsSubsystemStackset = "stackset"
 	metricsSubsystemStack    = "stack"
+	metricsSubsystemErrors   = "errors"
 )
 
 type MetricsReporter struct {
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
Is there no way to explicitly check whether a `watch` is active & syncing or not? Just an error counter might cause too many false positives.

linki:
We tried to find such a thing. This is the best we could come up with.

linki:
:+1:

njuettner:
👍

PR #204, opened 2020-02-26
Add a health check with last time of sync

linki:
This adds an extensible health check endpoint as well as an initial check for the last sync time.

The output can be seen at `http://127.0.0.1:7979/healthz?full=1`.

This particular check starts out unhealthy and switches to healthy for as long as the last sync interval isn't missed for too long (this code sets it to two intervals).

This can be used
* in a `readinessProbe` to ensure the initial sync succeeds, as well as
* in a `livenessProbe` to restart stackset-controller in case it gets stuck

### DIFF
--- a/cmd/stackset-controller/main.go
+++ b/cmd/stackset-controller/main.go
@@ -10,7 +10,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/heptiolabs/healthcheck"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


aermakov-zalando:
It'll still be healthy before the first `Run` starts, might want to add a dummy failed check when registering.

linki:
I moved registration of `/healthz` to after `startWatch()` is called. This should avoid any false positives during initial bootstrap.

aermakov-zalando:
Could you rebase on master? I've fixed the e2e deployment.

aermakov-zalando:
Otherwise LGTM.

linki:
:+1:

njuettner:
👍

PR #210, opened 2020-03-16
Update the CRDs with patched autogenerated versions

arjunrn:
I took the generated CRD from #182 and patched them with the missing fields from `master` and just committed the files only.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #211, opened 2020-03-17
Status should not be required in stacks

mikkeloscar:
Avoid having Status required as it breaks when deploying a resource without the status defined as the stackset controller does when initially creating stacks.


mikkeloscar:
:+1:

arjunrn:
👍

PR #213, opened 2020-03-30
CRD: don't require StackSet.status

thilp:
Just like in #211 for Stack, I don't think StackSet.status should be required on creation.


mikkeloscar:
:+1:

thilp:
:+1:

Issue #31, opened 2018-08-23
Enrich "status" section of StackSets with more information


linki:
The `status` section of StackSets contains useful information about the readiness of the StackSet. However, there's more information that could be exposed there so that tools probing for its readiness would have a single, reliable point to look at.

Useful inforamtion includes but isn't limited to:
- [ ] availability of the generated Ingress' public DNS name

PR #215, opened 2020-04-29
Increase event QPS as a quick workaround for deployment probers

aermakov-zalando:
Since https://github.com/zalando-incubator/stackset-controller/issues/31 is not solved, our probers still have to rely on monitoring `Failed…` events to figure out if the stackset is healthy. We currently determine if the events are 'old enough' by using a 1 minute range, but since `EventBroadcaster` has some fancy QPS logic inside, it will start dropping event updates at some point. Let's increase it even further for now as a quick hack.


aryszka:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
Considering that this is more than 5 years old and not something I am aware have been raised as an issue. I would not merge it unless we have a clear and current need.

mikkeloscar:
Closing this for now

PR #218, opened 2020-05-20
Update annotations in per-stack ingresses

aermakov-zalando:
Per-stack ingresses should be updated when the annotations in `stackset.ingress.metadata` change as well, otherwise filter/predicate changes won't be propagated. We should probably handle label changes as well, but doing this correctly is way too annoying. :/

### DIFF
--- a/controller/stack_resources.go
+++ b/controller/stack_resources.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/api/autoscaling/v2beta2"
 	apiv1 "k8s.io/api/core/v1"
 	networking "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -213,7 +214,7 @@ func (c *StackSetController) ReconcileStackIngress(ctx context.Context, stack *z
 	}
 
 	// Check if we need to update the Ingress
-	if core.IsResourceUpToDate(stack, existing.ObjectMeta) {
+	if core.IsResourceUpToDate(stack, existing.ObjectMeta) && equality.Semantic.DeepEqual(ingress.Annotations, existing.Annotations) {
 		return nil
 	}
 
--- a/controller/stack_resources_test.go
+++ b/controller/stack_resources_test.go
@@ -625,6 +625,9 @@ func TestReconcileStackIngress(t *testing.T) {
 		},
 	}
 
+	baseTestStackOwnedUpdatedAnnotations := *baseTestStackOwned.DeepCopy()
+	baseTestStackOwnedUpdatedAnnotations.Annotations["example"] = "updated"
+
 	for _, tc := range []struct {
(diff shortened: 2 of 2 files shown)


controller/stack_resources.go
```diff
@@ -210,7 +211,7 @@ func (c *StackSetController) ReconcileStackIngress(stack *zv1.Stack, existing *e
 	}
 
 	// Check if we need to update the Ingress
-	if core.IsResourceUpToDate(stack, existing.ObjectMeta) {
+	if core.IsResourceUpToDate(stack, existing.ObjectMeta) && equality.Semantic.DeepEqual(ingress.Annotations, existing.Annotations) {
```
aryszka, line 214: will this check for annotations like `deployment-time` or `kubectl.kubernetes.io/last-applied-configuration`, too?

aermakov-zalando: These are labels, not annotations.

aryszka: `deployment-id` is a label, but `deployment-time` is an annotation (-time). Both are there as annotations, if we just read the ingress object from the API.

aermakov-zalando: I'm still not sure about the question. The per-stack ingresses should in theory be modifiable by the users (in case they want to add a label or an annotation), but this will complicate things a lot and it's not clear if this is actually a valid use case (for example how do we resolve conflicts?). They also don't have any labels or annotations except those propagated from the stack or from the stackset's ingress object, because we don't `apply` them in stackset controller. Even if `DeepEqual` returns false, we'll just overwrite whatever is in there with the expected value and won't do anything to it on the next iteration.

mikkeloscar:
Can you elaborate on the use case you have where this would be helpful?

On the one hand stacks are created based on the StackSet state at the creation time and after that not updated when the stackset is upated. So based on this I would argue that semantically the annotations shouldn't be updated. However on the other side if they're not updated you don't really have a way to update them on a stack because editing the stack won't help.
So I think it would be helpful just to understand the use case and think about if it can cause any regression for other use cases.

aermakov-zalando:
@mikkeloscar Ingress configuration is global, it doesn't even exist in the stack's `spec`, so it should also be updated when the source of truth in the stackset object is updated. Right now, for example, if you make a mistake in the predicates or filters, you'll have to delete the ingresses even if you fix it in the stackset.

aermakov-zalando:
Although maybe I should change it even further. Even though ingress settings are global, they won't be updated unless the users also do unrelated changes in the stack, which is not very intuitive.

controller/stack_resources.go
aryszka: we discussed this in chat: I was just simply not aware that the source of the last-applied-configuration is kubectl apply. Thanks for the explanation @aermakov-zalando !

gargravarr:
Ingress settings are not global anymore since StackSet controller implements ingress versioning since https://github.com/zalando-incubator/stackset-controller/pull/495 We can close this.

Issue #222, opened 2020-06-15
metadata from stackset service won't be propagated


szuecs:
stackset
```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
....
  name: myapp
  namespace: shadow-test
spec:
...
  stackTemplate:
    metadata:
      creationTimestamp: null
    spec:
      replicas: 3
      service:
        metadata:
          annotations:
            zalando.org/api-usage-monitoring-tag: shadow-test   <--- not propagated to service
          creationTimestamp: null
...
```

service:
```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:    <--- missing
    stackset-controller.zalando.org/stack-generation: "2"
  creationTimestamp: "2020-06-11T16:17:50Z"
  labels:
    application: myapp
    shadow_test_version: dl1
    stack-version: master-9
    stackset: myapp-9
  name: myapp-9
  namespace: shadow-test
```

Code generating the service with annotations: https://github.com/zalando-incubator/stackset-controller/blob/2baddca617e2b76e34976357765206280cfd382e/pkg/core/stack_resources.go#L63-L71

PR #224, opened 2020-06-18
Add service metadata from stack spec to service generated by stack

chandu188:
Fixes https://github.com/zalando-incubator/stackset-controller/issues/222

### DIFF
--- a/pkg/core/stack_resources.go
+++ b/pkg/core/stack_resources.go
@@ -248,8 +248,14 @@ func (sc *StackContainer) GenerateService() (*v1.Service, error) {
 		return nil, err
 	}
 
+	metaObj := sc.resourceMeta()
+	stackSpec := sc.Stack.Spec
+	if stackSpec.Service != nil {
+		metaObj.Annotations = mergeLabels(metaObj.Annotations, stackSpec.Service.Annotations)
+		metaObj.Labels = mergeLabels(metaObj.Labels, stackSpec.Service.Labels)
+	}
 	return &v1.Service{
-		ObjectMeta: sc.resourceMeta(),
+		ObjectMeta: metaObj,
 		Spec: v1.ServiceSpec{
 			Selector: sc.selector(),
 			Type:     v1.ServiceTypeClusterIP,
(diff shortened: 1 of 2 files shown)


chandu188:
@szuecs Could you give me some feedback on the PR?

pkg/core/stack_resources.go
```diff
@@ -248,8 +248,14 @@ func (sc *StackContainer) GenerateService() (*v1.Service, error) {
 		return nil, err
 	}
 
+	metaObj := sc.resourceMeta()
+	stackSpec := sc.Stack.Spec
+	if stackSpec.Service != nil {
+		metaObj.Annotations = mergeLabels(metaObj.Annotations, stackSpec.Service.Annotations)
+		metaObj.Labels = mergeLabels(metaObj.Labels, stackSpec.Service.Labels)
```
mikkeloscar, line 255: On ingress we only insert annotations, maybe we should do the same here. I don't see a need for inserting labels so let's keep it out until we have a use case for it IMO.

cjbooms: You sure about this @mikkeloscar I see labels being propagated onto StackSet owned `Ingress` resources. EG:
```
kind: Ingress
metadata:
  annotations:
...
  labels:
    application: rules-evaluator
    deployment-id: d-uossw3sit9rfgwhj776zv84ky
    pipeline-id: l-28ykzm3xih43dne4518ubhjs3i
    stackset: rules-evaluator
```

cjbooms: Oh nevermind, I see the difference. You're referring to labels defined on the Service itself

chandu188:
@mikkeloscar removed labels as per the suggestion !!

mikkeloscar:
:+1:

cjbooms:
Any ETA for when this will be merged? I'm hoping to benchmark a new API for Pop and would like my Service annotation to get picked up so I can differentiate the metrics:
```
  stackTemplate:
    spec:
      service:
        metadata:
          annotations:
            zalando.org/api-usage-monitoring-tag: "load-test"
```

szuecs:
:+1:

PR #225, opened 2020-06-26
Use Networking API instead of Extensions for Ingress

mikkeloscar:
Updates all cases of Ingress interaction to use the newer `networking.k8s.io/v1beta1` apiVersion instead of `extensions/v1beta1`

This also updates the Autoscaling config to create HPA with ingress type pointing to the `networking.k8s.io/v1beta1` api. This is supported since https://github.com/zalando-incubator/kube-metrics-adapter/releases/tag/v0.1.5

PR #232, opened 2020-07-21
Fix ingress api-version

mikkeloscar:
Follow up to #225 

This fixes the apiversion for ingress such that it will work with kube-metrics-adapter. Without the HPA will report errors like:

```
- lastTransitionTime: "2020-07-21T15:04:57Z"
    message: 'the HPA was unable to compute the replica count: unable to get metric
      requests-per-second,x: Ingress on default x/unable
      to fetch metrics from custom metrics API: unable to map kind Ingress.networking
      to resource: no matches for kind "Ingress" in group "networking"'
    reason: FailedGetObjectMetric
    status: "False"
    type: ScalingActive
```


mikkeloscar:
:+1:

jrake-revelant:
👍

PR #235, opened 2020-08-14
Fix CRD validation

mikkeloscar:
Fixes CRD validation the following way:
* Remove `x-kubernetes-list-map-keys:` which doesn't work: https://github.com/kubernetes-sigs/controller-tools/issues/444
* Add status validation to StackSet CRD validation
* Set `preserveUnknownFields: false` such that unknown fields gives a validation error.

### DIFF
--- a/docs/stack_crd.yaml
+++ b/docs/stack_crd.yaml
@@ -1,45 +1,59 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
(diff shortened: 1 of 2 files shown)


szuecs:
Should the change be visible in VERSION?

mikkeloscar:
@szuecs I'm not sure this requires a bump in the version. We plan to merge #194 soon anyway which will do a version bump so it could be part of that IMO.

I will close this issue in favor of automatically generating crd from source which is now working great: #182

PR #237, opened 2020-08-17
Make setting controller-id optional

mikkeloscar:
Setting the `CONTROLLER_ID` is useful for running e2e tests with a custom stackset-controller. But for using the default controller in the cluster it should be possible to not set the ID at all. This we want when running stackset e2e in kubernetes-on-aws.


mikkeloscar:
:+1:

linki:
:+1:

PR #238, opened 2020-08-19
Status should not be required in CRD

mikkeloscar:
Was previously fixed in #211 but forgot to remove it in #182


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #240, opened 2020-08-26
Fix defaults for autoscaling behavior

mikkeloscar:
When creating a StackSet where autoscaling behavior was specified but where `StabilizationWindowSeconds` was not set it would result in the following error when stackset controller attempted to create a stack:

```
Unable to create stack: Stack.zalando.org \"thisisfine-main-mlarsen-pr-1-5\" is invalid: spec.autoscaler.behavior.scaleDown.stabilizationWindowSeconds: Invalid value: \"null\": spec.autoscaler.behavior.scaleDown.stabilizationWindowSeconds in body must be of type integer: \"null\""
```

This happens because the CRD validation expects an integer value to be set or the field to not be included at all. ~As a work around we can set the defaults which would otherwise be applied by the APIServer before we attempt to create the stack. This will allow users to omit the `StabilizationWindowSeconds` and rely on the defaults when desired.~

As a workaround we introduce our own type for `HorizontalPodAutoscalerBehavior` where optional fields are correctly annotationed with `omitempty` (see #245). Hopefully we can also upstream a fix and use the normal types in the future, but for now we have to live with our own copy.

Additionally this PR also:
* Adds test for this specific case
* Moves the metric types to `types.go` similar to how it's done for core Kubernetes API resource definitions. This has the benefit of making the metric type constant names available to be used from any package and it also expands the CRD validation spec with some more description.


muaazsaleem:
👍

mikkeloscar:
:-1:

mikkeloscar:
Seems to not work as expected.

mikkeloscar:
Changed it to add the work-around in the right place e.g. one stack creation rather than on HPA creation.

muaazsaleem:
Was able to re-produce the test failure locally in `stups-test` but not in `platform-infrastructure-test`. This leads me to believe that it's CRD related.

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #241, opened 2020-08-26
Include Ingress annotation in e2e test

mikkeloscar:
We had an issue where ingress annotations where dropped when we enabled StackSet CRD validation. Include setting an annotation in the e2e tests.

Also tests setting annotation on the optional service definition.

### DIFF
--- a/cmd/e2e/basic_test.go
+++ b/cmd/e2e/basic_test.go
@@ -15,24 +15,29 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	userTestAnnotation = "user-test-annotation"
+)
+
 var (
 	pathType = v1beta1.PathTypeImplementationSpecific
 )
 
 type TestStacksetSpecFactory struct {
-	stacksetName    string
-	hpa             bool
-	ingress         bool
-	externalIngress bool
-	limit           int32
-	scaleDownTTL    int64
-	replicas        int32
-	hpaMaxReplicas  int32
-	hpaMinReplicas  int32
-	autoscaler      bool
-	maxSurge        int
-	maxUnavailable  int
-	metrics         []zv1.AutoscalerMetrics
+	stacksetName       string
+	hpa                bool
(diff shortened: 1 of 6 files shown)


mikkeloscar:
Depends on #242

cmd/e2e/basic_test.go
```diff
@@ -206,6 +221,7 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 	// Verify service
 	service, err := waitForService(t, stack.Name)
 	require.NoError(t, err)
+	require.Contains(t, service.Annotations, userTestAnnotation)
```
aermakov-zalando, line 224: Test for equality on the map, maybe? This way we don't accidentally add something in there, however unlikely this is.

aermakov-zalando: Might also be a good idea to add an explicit test for changing these, since this is something that's often done for at least the ingresses.

mikkeloscar: Currently we add `stackset-controller.zalando.org/stack-generation` into the resources which is why I just wanted to check for the annotation we add, otherwise the test becomes very specific to the implementation. WDYT?

I have added a test for updating ingress.

aermakov-zalando: I see! If this is too annoying and we test for the value anyway then let's keep it like this.

I'm not seeing the updating test, but maybe I'm just missing it. What I've meant is a test that would verify that changing the annotation value (or adding a new one) in the stackset correctly updates those on the ingress object as well.

mikkeloscar: > I'm not seeing the updating test, but maybe I'm just missing it.

Sorry I was commenting it out because some tests were failing. It's now added as `TestStacksetUpdateIngress` this creates first an ingress without annotation, and then it adds the annotation when updating the ingress.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #242, opened 2020-08-26
Fix ObjectMeta CRD validation

mikkeloscar:
As mentioned in #241 the CRD validation for the ObjectMeta fields which we used in various places doesn't work because the schema is just `object` for the whole `metadata` field so anything below is dropped.

This PR aims to fix the problem in two ways.

1. Introduce a new type `EmbeddedObjectMetaWithAnnotations` which is a slimmed down version of `metav1.ObjectMeta` that only has `Annotations` as a field. This has the bonus of only exposing the field which is actually used by the stackset controller for various sub resources e.g. setting a custom annotation on the ingress or service resources (`Labels` are not used, so no need to allow users to set them for instance).
2. ~It uses a patched version of https://github.com/kubernetes-sigs/controller-tools based on this PR: https://github.com/kubernetes-sigs/controller-tools/pull/395. (I have made a fork and are pulling it from there: https://github.com/mikkeloscar/controller-tools/commit/54e4525f9990a5510b454b7e1e132b6669b2cee9, I'm open for suggestions on how to better manage this).~
2. Introduce a new type `EmbeddedObjectMeta` which is a slimmed down version of `metav1.ObjectMeta` that has `Labels` and `Annotations`. This is used for a newly defined `PodTemplateSpec` type which is similar to `v1.PodTemplateSpec` but has `EmbeddedObjectMeta` instead of `metav1.ObjectMeta`.

Additionally it drops `metav1.ObjectMeta` from the `HorizontalPodAutoscaler` field. This is not in use in the code, so it doesn't make sense to support users setting it. We may want to add it in the future as annotations may be needed for interacting with https://github.com/zalando-incubator/kube-metrics-adapter but we have the `Autoscaler` field supporting this already. I would like to refactor the `HorizontalPodAutoscaler` field anyway in the future.


aermakov-zalando:
What about `labels`? Do we support those?

muaazsaleem:
> Labels are not used, so no need to allow users to set them for instance
@aermakov-zalando ^ from the issue description

pkg/apis/zalando.org/v1/types.go
```diff
@@ -53,13 +53,26 @@ type StackSetSpec struct {
 	Traffic []*DesiredTraffic `json:"traffic,omitempty"`
 }
 
+// EmbededObjectMetaWithAnnotations defines the metadata which can be attached
+// to a resource. It's a slimmed down version of metav1.ObjectMeta only
+// containing annotations.
+// +k8s:deepcopy-gen=true
+type EmbededObjectMetaWithAnnotations struct {
```
aryszka, line 60: embedded, double-d

mikkeloscar: Thanks, fixed!

aryszka:
:+1:

mikkeloscar:
Deployments are failing with this in place:

image


Need to adjust the ObjectMeta of the PodTemplate.

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #245, opened 2020-08-31
Use custom type for HPA behavior

mikkeloscar:
Upstream definition for [`StabilizationWindowSeconds`](https://github.com/kubernetes/kubernetes/blob/9f2892aab98fe339f3bd70e3c470144299398ace/staging/src/k8s.io/api/autoscaling/v2beta2/types.go#L170) does not define `omitempty` even though the field is optional. This causes a problem when used in combination with our own types for `StackSet` and `Stack` because validation will fail when trying to submit a resource where the `StabilizationWindowSeconds` has not been defined:

```
StackSet.zalando.org "hpa-behavior" is invalid: spec.stackTemplate.spec.horizontalPodAutoscaler.behavior.scaleDown.stabilizationWindowSeconds: Invalid value: "null": spec.stackTemplate.spec.horizontalPodAutoscaler.behavior.scaleDown.stabilizationWindowSeconds in body must be of type integer: "null"
```

To work around this we instead define our own types for `HorizontalPodAutoscalerBehavior` and correctly mark the field `omitempty`.
This makes it a bit annoying to convert between the different types, but avoids us having to set arbitrary defaults as first attempted as the fix in #240


muaazsaleem:
👍

mikkeloscar:
:+1:

PR #252, opened 2020-09-16
Update CRDs to v1

mikkeloscar:
Update the generated CRD manifests to `apiextensions.k8s.io/v1`

Additionally:
* Cleanup the Makefile a bit
* Update `sigs.k8s.io/controller-tools` to latest commit which includes fixes that allows us to drop a work around for `x-kubernetes-list-map-keys` fields.
* Define enum for `AutoscalerMetricType`


mikkeloscar:
:+1:

Makefile
```diff
@@ -32,20 +31,13 @@ check: $(GENERATED)
 	go mod download
 	golangci-lint run --timeout=2m ./...
 
-$(GENERATED): $(CRD_TYPE_SOURCE)
+$(GENERATED): go.mod $(CRD_TYPE_SOURCE)
```
muaazsaleem, line 34: curious what adding `go.mod` here does?

Makefile
```diff
@@ -32,20 +31,13 @@ check: $(GENERATED)
...
 	./hack/update-codegen.sh
 
-$(CONTROLLER_GEN):
-	mkdir -p build
-	GOBIN=$(shell pwd)/build go install sigs.k8s.io/controller-tools/cmd/controller-gen
```
muaazsaleem, line 40: I guess `go install` wasn't necessary in this case because we didn't want to update our `${GOPATH}/bin` right?

Makefile
mikkeloscar: Changes to `go.mod` will now trigger re-generation/build which is often desirable i.e. if you update the generation tool version.

cmd/e2e/test_utils.go
```diff
@@ -44,6 +45,7 @@ var (
 				Ports: []corev1.ContainerPort{
 					{
 						ContainerPort: 80,
+						Protocol:      v1.ProtocolTCP,
```
muaazsaleem, line 48: Just for my info, is specifying the protocol necessary with `v1` or does this have some other effect?

Makefile
mikkeloscar: Right, we can just run `go run sigs.k8s.io/controller-tools/cmd/controller-gen` when needed instead of first installing the binary. So this is just to skip an unneeded extra step.

cmd/e2e/test_utils.go
mikkeloscar: With v1 the generated CRD schema sets the default for Protocol. We have to add it to our expected struct as we now get it included when we read the resource from the server.

muaazsaleem: Makes sense. I guess the tests would fail otherwise.

muaazsaleem:
👍

PR #257, opened 2020-09-18
Sort RG backends to have stable RG generation

mikkeloscar:
Sort the RouteGroup backends and default backends by name when generating RouteGroups to ensure that the generated resource doesn't change unless the input changes. Without this the controller may continuously update the resources without any functional change but emitting a lot of useless update events doing so.


muaazsaleem:
👍

mikkeloscar:
:+1:

PR #264, opened 2020-10-31
Add support for multiple cluster domains

mikkeloscar:
This extends the `--cluster-domain` flag introduced in #202 to represent a list of cluster domains (define the flag multiple times to define more domains).
This allows generating per stack ingresses/routegroups for multiple types of cluster domains i.e. for a public name like `example.org` and for an internal name like `ingress.cluster.local` as defined in the Zalando Kubernetes infrastructure.

The per stack ingress/rouegroup will only contain hostnames for cluster domains which are defined on the stackset. I.e. if the clusters domains are defined as: `[example.org,ingress.cluster.local]` and a stackset defines only `app.ingress.cluster.local`, then the per stack ingress/routegroup will only have `<stack>.ingress.cluster.local` defined and vice versa.

### DIFF
--- a/cmd/e2e/README.md
+++ b/cmd/e2e/README.md
@@ -4,8 +4,9 @@ The following environment variables should be set:
 
 1. `E2E_NAMESPACE` is the namespace where the tests should be run.
 2. `CLUSTER_DOMAIN` is the DNS domain managed in which ingresses should be created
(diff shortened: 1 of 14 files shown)


jonathanbeber (approved):

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #268, opened 2020-11-10
Look at the RouteGroup/Ingress UpdateTimestamp rather than CreationTimestamp

muaazsaleem:
The HTTP resources (Ingresses and RouteGroups) migration, introduced on
the [#251][0], checks for the `CreationTimestamp` attribute of the
resources. When the migration follows the path of:

1. adding a new resource (does not matter if it's an Ingress or
   RouteGroup) with a new hostname for migration and testing purposes;
2. validating the configuration of the new resource.
3. dropping the old resource in favor of the new one. For that the new
   resource receives the actual hostname.

the controller does not respect the [`IngressSourceSwitchTTL`][1]
duration as expected. During the 3rd step, the `CreationTimestamp`
attribute of the new resource tends to be greater than the
`IngressSourceSwitchTTL` configuration, due to the time spent in the 2nd
step. That leads to the almost instantaneous deletion of the old
resource, without considering the configured TTL for routes propagation.

This commit introduces a new annotation to the above-mentioned
resources: `stackset-controller.zalando.org/updated-timestamp`. This
annotation is responsible for tracking the last time the controller
updated a resource. Also, it changes the migration process introduced
on the [#251][0] to compare `IngressSourceSwitchTTL` with the newly
added annotation instead of the `CreationTimestamp` attribute. This way,
the change of hostname on the new resource, during the 3rd step, resets
the timer to delete the old resource and guarantees route propagation.

Also, it adds a new e2e test file. This new test create the above
mentioned migration process and verifies that the controller delete the
old resources after the `IngressSourceSwitchTTL` duration.

As a minor change, this commit also updates the `run_e2e.sh` script,
allowing one to define the environment variable `TEST_NAME`. It will set
the maximum parallelism to 1 and run the given test name.

[0]: #251
[1]: https://github.com/zalando-incubator/stackset-controller/blob/e72232e799e5ff8ba5b51d29dc5ddff153907162/cmd/stackset-controller/main.go#L42

### DIFF
--- a/cmd/e2e/ingress_source_switch_test.go
+++ b/cmd/e2e/ingress_source_switch_test.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	v1 "github.com/szuecs/routegroup-client/apis/zalando.org/v1"
+	"github.com/zalando-incubator/stackset-controller/controller"
+)
+
+const (
+	// The e2e env IngressSourceSwitchTTL
+	IngressSourceSwitchTTL = time.Minute
+)
+
+func TestIngressSourceSwitch(t *testing.T) {
+	t.Parallel()
+
+	stacksetName := "ingress-source-switch-stackset"
+	firstVersion := "v1"
+
+	// create stackset with ingress
+	factory := NewTestStacksetSpecFactory(stacksetName).Ingress(nil)
+	spec := factory.Create(firstVersion)
+	err := createStackSet(stacksetName, 0, spec)
+	require.NoError(t, err)
+	ingress, err := waitForIngress(t, stacksetName)
+	require.NoError(t, err)
+	require.Contains(t, ingress.Annotations, controller.ControllerLastUpdatedAnnotationKey)
+	require.NotEqual(t, "", ingress.Annotations[controller.ControllerLastUpdatedAnnotationKey])
+
+	// update stackset adding a routegroup
+	updatedVersion := "v2"
+	spec = factory.RouteGroup().Create(updatedVersion)
+	err = updateStackset(stacksetName, spec)
+	require.NoError(t, err)
+	rg, err := waitForRouteGroup(t, stacksetName)
+	require.NoError(t, err)
+	firstRgUpdatedTimestamp := ingress.Annotations[controller.ControllerLastUpdatedAnnotationKey]
+	require.Contains(t, rg.Annotations, controller.ControllerLastUpdatedAnnotationKey)
+	require.NotEqual(t, "", firstRgUpdatedTimestamp)
+	secondIngress, err := waitForIngress(t, stacksetName)
+	require.Equal(t, secondIngress.Annotations[controller.ControllerLastUpdatedAnnotationKey], ingress.Annotations[controller.ControllerLastUpdatedAnnotationKey])
+	require.NoError(t, err)
+
+	// update the routegroup and delete the ingress
+	lastVersion := "v3"
+	lastSpec := spec
+	lastSpec.StackTemplate.Spec.Version = lastVersion
+	lastSpec.RouteGroup.AdditionalBackends = []v1.RouteGroupBackend{{Name: "shunt", Type: v1.ShuntRouteGroupBackend}}
+	lastSpec.Ingress = nil
+	err = updateStackset(stacksetName, lastSpec)
+	require.NoError(t, err)
+	lastRg, err := waitForUpdatedRouteGroup(t, rg.Name, rg.Annotations[controller.ControllerLastUpdatedAnnotationKey])
+	require.NoError(t, err)
+	require.NotEqual(t, lastRg.Annotations[controller.ControllerLastUpdatedAnnotationKey], firstRgUpdatedTimestamp)
+	// If the ingress is not deleted right away, then, it respects the
+	// TTL.
+	err = resourceDeleted(t, "ingress", stacksetName, ingressInterface()).await()
+	require.Error(t, err)
+
+	// make sure the ingress is finally deleted after twice the TTL
+	a := resourceDeleted(t, "ingress", stacksetName, ingressInterface())
+	a.timeout += IngressSourceSwitchTTL * 2
+	err = a.await()
+	require.NoError(t, err)
+}
+
+func TestIngressToRouteGroupSwitch(t *testing.T) {
+	t.Parallel()
+
+	stacksetName := "ingress-to-routegroup-switch"
+	firstVersion := "v1"
+
+	// create stackset with ingress and routegroup
+	factory := NewTestStacksetSpecFactory(stacksetName).Ingress(nil).RouteGroup()
+	spec := factory.Create(firstVersion)
+	err := createStackSet(stacksetName, 0, spec)
+	require.NoError(t, err)
+	rg, err := waitForRouteGroup(t, stacksetName)
+	require.NoError(t, err)
+
+	// Wait the IngressSourceSwitchTTL to make sure the ingress is
+	// created a time long enough ago.
+	time.Sleep(IngressSourceSwitchTTL)
+
+	// update the routegroup and delete the ingress
+	lastVersion := "v2"
+	lastSpec := spec
+	lastSpec.StackTemplate.Spec.Version = lastVersion
+	lastSpec.RouteGroup.AdditionalBackends = []v1.RouteGroupBackend{{Name: "shunt", Type: v1.ShuntRouteGroupBackend}}
+	lastSpec.Ingress = nil
+	err = updateStackset(stacksetName, lastSpec)
+	require.NoError(t, err)
+	_, err = waitForUpdatedRouteGroup(t, rg.Name, rg.Annotations[controller.ControllerLastUpdatedAnnotationKey])
+	require.NoError(t, err)
+
+	// make sure ingress is not deleted before IngressSourceSwitchTTL
+	err = resourceDeleted(t, "ingress", stacksetName, ingressInterface()).await()
+	require.Error(t, err)
+}
+
+func TestRouteGroupToIngressSwitch(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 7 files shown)


cmd/e2e/test_utils.go
```diff
@@ -438,6 +440,34 @@ func waitForRouteGroup(t *testing.T, name string) (*rgv1.RouteGroup, error) {
 	return routegroupInterface().Get(context.Background(), name, metav1.GetOptions{})
 }
 
+func waitForUpdatedRouteGroup(t *testing.T, name string, oldTimestamp string) (*rgv1.RouteGroup, error) {
```
muaazsaleem, line 443: What if we create a single `func` i.e `waitForUpdatedResource` and pass it a `resourceVersion`. This way we could re-use the `func` for `Ingress`, `RouteGroup` and all other resources?

We could still check for that `updateTimestamp` was updated in the e2e but that's already tested in the controller integration tests anyways. WDYT?

controller/stackset.go
```diff
@@ -34,9 +34,10 @@ import (
 )
 
 const (
-	PrescaleStacksAnnotationKey               = "alpha.stackset-controller.zalando.org/prescale-stacks"
-	ResetHPAMinReplicasDelayAnnotationKey     = "alpha.stackset-controller.zalando.org/reset-hpa-min-replicas-delay"
-	StacksetControllerControllerAnnotationKey = "stackset-controller.zalando.org/controller"
+	PrescaleStacksAnnotationKey                    = "alpha.stackset-controller.zalando.org/prescale-stacks"
+	ResetHPAMinReplicasDelayAnnotationKey          = "alpha.stackset-controller.zalando.org/reset-hpa-min-replicas-delay"
+	StacksetControllerControllerAnnotationKey      = "stackset-controller.zalando.org/controller"
+	StacksetControllerUpdateTimestampAnnotationkey = "stackset-controller.zalando.org/updated-timestamp"
```
muaazsaleem, line 40: nit: WDYT abt renaming this to:

```
ControllerLastUpdatedAnnotationkey = "stackset-controller.zalando.org/last-updated"
```
The timestamp part is will be obvious with the value anyways.

cmd/e2e/ingress_source_switch_test.go
```diff
@@ -0,0 +1,133 @@
...
+	lastSpec.RouteGroup.AdditionalBackends = []v1.RouteGroupBackend{{Name: "shunt", Type: v1.ShuntRouteGroupBackend}}
+	lastSpec.Ingress = nil
+	err = updateStackset(stacksetName, lastSpec)
+	require.NoError(t, err)
+	_, err = waitForUpdatedRouteGroup(t, rg.Name, rg.Annotations[updatedTimeStampAnnotationKey])
+	require.NoError(t, err)
+
+	// make sure ingress is not deleted before IngressSourceSwitchTTL
+	_, err = waitForIngress(t, stacksetName)
+	require.NoError(t, err)
```
muaazsaleem, line 98: When trying out this  scenario manually, this is failing for me. Not sure yet why this test is passing.

muaazsaleem: Note that `ReconcileStackSetIngress` is still getting the "old" RouteGroup. So it won't accurately know when `RouteGroup` was last updated.

https://github.com/zalando-incubator/stackset-controller/blob/5292825618e79dce55753f6d10d0faeaed809492/controller/stackset.go#L854

cmd/e2e/test_utils.go
jonathanbeber: Hey, @muaazsaleem I agree with you here and that was my first attempt. The result was a few flaky tests. If you recall, we were also comparing the resource version [here](https://github.com/zalando-incubator/stackset-controller/blob/5292825618e79dce55753f6d10d0faeaed809492/cmd/e2e/ingress_source_switch_test.go#L43), but it used to fail from time to time. My first hypothesis is that other "controllers" also operate these resources (in our case skipper) and it affects resource version in a manner we can't control. I'd be happy to invest more time on that, I'm just not sure that should be the focus for this PR. What are your thoughts?

muaazsaleem: Oh that's totally possible. The controller annotation makes more sense then.

cmd/e2e/ingress_source_switch_test.go
jonathanbeber: We fixed the e2e test to get it. We were testing that the ingress was still there just once. Now it waits until the `waiter` timeout to guarantee that the ingress is not deleted.
 
We solved this issue setting a new func `ReconcileStackSetIngressSources` that better represents the strong relationship between ingresses and route groups at this moment. We also unified the `ReconcileStackSetIngress`' and the `ReconcileStackSetRouteGroups`' unit tests.

controller/stackset.go
```diff
@@ -651,161 +658,255 @@ func (c *StackSetController) CleanupOldStacks(ctx context.Context, ssc *core.Sta
...
+		apiv1.EventTypeNormal,
+		"DeletedIngress",
+		"Deleted Ingress %s",
+		existing.Namespace)
+	return nil
+}
 
-	// RouteGroup removed
+// ReconcileStackSetRouteGroup reconciles the RouteGroup but never deletes it, it returns the existing/new RouteGroup
+func (c *StackSetController) ReconcileStackSetRouteGroup(ctx context.Context, stackset *zv1.StackSet, existing *rgv1.RouteGroup, ingress *networking.Ingress, rg *rgv1.RouteGroup) (*rgv1.RouteGroup, error) {
```
muaazsaleem, line 760: I guess we could rename this funcs `AddUpdateStackSetRouteGroup` in a future PR

controller/stackset.go
```diff
@@ -651,161 +658,255 @@ func (c *StackSetController) CleanupOldStacks(ctx context.Context, ssc *core.Sta
...
+		"DeletedRouteGroup",
+		"Deleted RouteGroup %s",
+		rg.Namespace)
 	return nil
 }
 
-func (c *StackSetController) ReconcileStackSetResources(ctx context.Context, ssc *core.StackSetContainer) error {
-	// opt-out ingress creation in case we have an external entity creating ingress
-	err := c.ReconcileStackSetIngress(ctx, ssc.StackSet, ssc.Ingress, ssc.RouteGroup, ssc.GenerateIngress)
+func (c *StackSetController) ReconcileStackSetIngresSources(
```
muaazsaleem, line 849: nit: `ReconcileStackSetIngresSources ` -> `ReconcileStackSetIngressSources `

controller/stackset_test.go
```diff
@@ -487,404 +511,798 @@ func TestReconcileStackSetIngress(t *testing.T) {
...
+			}
+			if tc.existingRg != nil {
+				err = env.CreateRouteGroups(context.Background(), []rgv1.RouteGroup{*tc.existingRg})
 				require.NoError(t, err)
 			}
 
-			err = env.controller.ReconcileStackSetRouteGroup(context.Background(), &stackset, tc.existing, tc.ingress, func() (*rgv1.RouteGroup, error) {
-				return tc.updated, nil
-			})
+			// TODO: test disabling routegroup support
```
muaazsaleem, line 1267: @jonathanbeber WDYT abt this? I guess it does make sense to have a test to make sure that disabling rg support actually works. How were you thinking of doing it?

muaazsaleem:
👍

controller/stackset_test.go
jonathanbeber: I believe we can just create a test with a declared route group and making sure it's not created. Also, the ingress should be deleted without waiting for the TTL if the route group support is disabled. Let's add these tests.

controller/stackset_test.go
```diff
@@ -449,8 +449,8 @@ func TestReconcileStackSetDesiredTraffic(t *testing.T) {
 	}
 }
 
-func TestReconcileStackSetIngress(t *testing.T) {
-	exampleRules := []networking.IngressRule{
+func TestReconcileStackSetIngresSources(t *testing.T) {
```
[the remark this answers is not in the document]
jonathanbeber: good one!

cmd/e2e/test_utils.go
```diff
@@ -35,6 +35,8 @@ const (
 
 	weightKindDesired weightKind = "zalando.org/stack-traffic-weights"
 	weightKindActual  weightKind = "zalando.org/backend-weights"
+
+	updatedTimeStampAnnotationKey = "stackset-controller.zalando.org/updated-timestamp"
```
mikkeloscar, line 39: nit: The `controller` package is already imported here so you can just use the const from there: `controller.ControllerLastUpdatedAnnotationKey`

controller/stackset.go
```diff
@@ -651,161 +658,255 @@ func (c *StackSetController) CleanupOldStacks(ctx context.Context, ssc *core.Sta
...
-		_, err := c.client.NetworkingV1beta1().Ingresses(ingress.Namespace).Create(ctx, ingress, metav1.CreateOptions{})
+		if ingress.Annotations == nil {
+			ingress.Annotations = make(map[string]string)
+		}
+		ingress.Annotations[ControllerLastUpdatedAnnotationKey] = c.now()
+
+		createdIng, err := c.client.NetworkingV1beta1().Ingresses(ingress.Namespace).Create(ctx, ingress, metav1.CreateOptions{})
 		if err != nil {
-			return err
+			return existing, err
```
mikkeloscar, line 678: nit: I think it's more clear to explicitly return `nil` in the error cases

jonathanbeber: Makes sense, in any case, the caller function check the err and just return, so returning existing have no effect. Will apply. Thanks.

controller/stackset_test.go
jonathanbeber: That's done, created two new test cases: `routegroup isn't created if its support is disabled` and `ingress is removed right away if RouteGroup support is disabled`.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #274, opened 2020-12-08
Switch to a compliant base image

aermakov-zalando:
I want to stop getting the emails. :(


aermakov-zalando:
👍

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.opensource.zalan.do/stups/alpine:latest
+FROM registry.opensource.zalan.do/library/alpine-3.12:3.12-20201117
```
mikkeloscar, line 1: can't we use `:latest` here? so we don't have to manually update in the future?

aermakov-zalando: Fixed!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #280, opened 2020-12-14
Make RG fields required

mikkeloscar:
Make relevant RouteGroup fields required in CRD spec.

Turns out if they're annotated with `omitempty` then they are treated as optional. We want these fields to be required.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #299, opened 2021-02-22
fix routegroup stack ttl handling erported by internal user

szuecs:

### DIFF
--- a/pkg/core/stackset.go
+++ b/pkg/core/stackset.go
@@ -111,10 +111,9 @@ func (ssc *StackSetContainer) MarkExpiredStacks() {
 	gcCandidates := make([]*StackContainer, 0, len(ssc.StackContainers))
 
 	for _, sc := range ssc.StackContainers {
-		// Stacks are considered for cleanup if we don't have an ingress or if the stack is scaled down because of inactivity
-		hasIngress := sc.ingressSpec != nil || ssc.StackSet.Spec.ExternalIngress != nil
-		hasRouteGroup := sc.routeGroupSpec != nil
-		if !(hasIngress || hasRouteGroup) || sc.ScaledDown() {
+		// Stacks are considered for cleanup if we don't have RouteGroup nor an ingress or if the stack is scaled down because of inactivity
+		hasIngress := sc.routeGroupSpec != nil || sc.ingressSpec != nil || ssc.StackSet.Spec.ExternalIngress != nil
+		if !hasIngress || sc.ScaledDown() {
 			gcCandidates = append(gcCandidates, sc)
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/core/stackset.go
```diff
@@ -113,7 +113,9 @@ func (ssc *StackSetContainer) MarkExpiredStacks() {
 	for _, sc := range ssc.StackContainers {
 		// Stacks are considered for cleanup if we don't have an ingress or if the stack is scaled down because of inactivity
 		hasIngress := sc.ingressSpec != nil || ssc.StackSet.Spec.ExternalIngress != nil
-		if !hasIngress || sc.ScaledDown() {
+		hasRouteGroup := sc.routeGroupSpec != nil
+		if !(hasIngress || hasRouteGroup) || sc.ScaledDown() {
+			//if !hasIngress || sc.ScaledDown() {
```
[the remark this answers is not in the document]
szuecs: thanks fixed, wanted to make sure the test fail :)

szuecs:
:+1:

pkg/core/stackset.go
```diff
@@ -113,7 +113,8 @@ func (ssc *StackSetContainer) MarkExpiredStacks() {
 	for _, sc := range ssc.StackContainers {
 		// Stacks are considered for cleanup if we don't have an ingress or if the stack is scaled down because of inactivity
 		hasIngress := sc.ingressSpec != nil || ssc.StackSet.Spec.ExternalIngress != nil
-		if !hasIngress || sc.ScaledDown() {
+		hasRouteGroup := sc.routeGroupSpec != nil
+		if !(hasIngress || hasRouteGroup) || sc.ScaledDown() {
```
aermakov-zalando, line 117: Nit: I'd move `sc.routeGroupSpec != nil` into the previous condition, just to avoid having to parse `!(foo||bar) || baz`.

szuecs: Not sure if I think it's more readable to put more into a var that would have a wrong name afterwards.

I thought "not has ingress nor has routegroup or scaled down" can be read easily

aermakov-zalando: We could rename the var, but feel free to ignore the suggestion.

szuecs: I see your point of reducing the code parse time.
Any suggestion to rename the var?

szuecs: I changed the comment to reflect the change and let it `hasIngress` maybe it's good enough as var name

szuecs:
:+1:

pkg/core/stackset.go
aermakov-zalando: No idea, I'd honestly keep it as-is because a route group is just a different kind of ingress.

aermakov-zalando:
👍

szuecs:
:+1:

PR #305, opened 2021-03-04
Hash the metric name

jonathanbeber:
This commit changes the way the `stack` label is set on the HPA to hash
the stackset name and its namespace instead of concatenating both to
avoid labels longer than 63 chars.


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #306, opened 2021-03-24
Support ContainerResource metric type in autoscaler

mikkeloscar:
Kubernetes v1.20 introduced an alpha feature for scaling based on individual container (cpu/memory) as opposed to scaling on the average across all containers in a pod: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#container-resource-metrics

This PR updates client-go to 1.20 and introduces optional support for specifying a container name on the `CPU` and `Memory` metric types of the `autoscaler` definition. If `container` is set to a non-empty value then the `ContainerResource` type will be used instead of the standard `Resource` type.

Additionally we can get rid of the custom copy of the HPA structs as v1.20 includes a fix for the struct definitions that we worked around before (https://github.com/kubernetes/kubernetes/pull/94372)

This is still backward compatible with v1.19 clusters if we just remove the new fields in the CRD so it's not possible to submit a StackSet with the new feature.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #315, opened 2021-04-01
Fix the panic on ZMON metrics

aermakov-zalando:
Changes:
 * Don't crash if a single stackset causes a panic. This should be fairly safe because stacksets are completely independent from each other, and it would allow the rest of the updates to succeed.
 * Correctly validate a missing ZMON metric.

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"runtime/debug"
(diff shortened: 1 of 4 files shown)


jonathanbeber:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #320, opened 2021-05-06
Add both resources to the 'all' category

aermakov-zalando:
This was apparently lost when we migrated to `controller-gen`. Additionally, fix the invocation since it just doesn't work at all without it.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #314, opened 2021-04-01
Update dependencies

jonathanbeber:
This commit is the output of `go get -u`. It also updates kubebuilder
which changes the way descriptions are managed on the generated CRDS.

PR #321, opened 2021-05-06
Downgrade Kubernetes dependencies to 1.19, regenerate CRDs

aermakov-zalando:
These were updated in https://github.com/zalando-incubator/stackset-controller/pull/314, but we don't support 1.20 in our clusters yet.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #328, opened 2021-06-01
Add a ScalingSchedule metrics for autoscaler

jonathanbeber:
This commit creates integration with the [`ScalingSchedule`
collectors][0] for autoscaler adding two new metric types:
`ScalingSchedule` and `ClusterScalingSchedule`. The metrics
configuration are pretty simple and receives just the object name.

[0]: https://github.com/zalando-incubator/kube-metrics-adapter#scalingschedule-collectors


pkg/core/autoscaler.go
```diff
@@ -292,6 +302,62 @@ func zmonMetric(metrics zv1.AutoscalerMetrics, stackName, namespace string) (*au
...
+		return nil, fmt.Errorf("average not specified")
+	}
+	average := metrics.Average.DeepCopy()
+
+	if metrics.ScalingSchedule == nil {
+		return nil, errMissingScalingScheduleDefinition
+	}
+
+	name := metrics.ScalingSchedule.Name
+	if name == "" {
```
mikkeloscar, line 316: This we could also catch already in the CRD validation with a pattern similar to what is used for validating `metadata.name`

jonathanbeber: We do, right? I didn't set this field as optional, so I assume it's done, but I find a weak link to do not validate here. If you think we could just drop it?

jonathanbeber: hmm, I got it now, it indeed allows me to set an empty string. I will add the validation. Apart from that, do you think we should/could drop this validation here?

jonathanbeber:
@mikkeloscar thanks for taking a look, I added a CRD pattern validation, the same used by the `metadata.name` as you suggested.

pkg/core/autoscaler.go
mikkeloscar: > so both might be fine.

Yes, both are fine, but it's a better user experience to catch it sooner.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #329, opened 2021-06-03
Allow customising per-stack ingresses/routegroups

aermakov-zalando:
This adds two new fields (`ingress`/`routegroup`) to the `stackTemplate` that allow for a better control over the per-stack ingresses and route-groups. These now can be disabled altogether (by setting `enabled: false`), use a different set of hostnames from the current autogenerated one, or different annotations.

This addresses a bunch of issues we've had with the per-stack ingresses:
 * not being possible to use the same stackset name in different namespaces (because the autogenerated hostname doesn't include the namespace)
 * extremely long deployment time for users that (for some reason) use a dedicated load balancer on their main ingress, since the annotation is copied to the per-stack ingresses even though it makes even less sense there


pkg/apis/zalando.org/v1/types.go
```diff
@@ -96,13 +96,38 @@ type EmbeddedObjectMeta struct {
...
 
+// +k8s:deepcopy-gen=true
+type StackIngressRouteGroupOverrides struct {
+	EmbeddedObjectMetaWithAnnotations `json:"metadata,omitempty"`
+
+	// Whether to enable per-stack ingresses or routegroups. Defaults to enabled if unset.
+	// +optional
+	Enabled *bool `json:"enabled"`
+
+	// Hostnames to use for the per-stack ingresses (or route groups). These must contain the special $stack_name$ token,
```
jonathanbeber, line 107: we could validate here if the token is present with a regex.

jonathanbeber: Also, I'm not sure if the pattern `$var_name$` is adapted in any other place, we could use something more familiar, like `${var_name}` or `{{{var_name}}}`

jonathanbeber (commented):
sounds good to me as an optional field. We will have to put a decent amount of work into documenting it but I think it will solve important issues. Thanks for the proposal.

pkg/apis/zalando.org/v1/types.go
aermakov-zalando: Those two would conflict with CDP, though, and I'm not sure if I want to subject the users to the escaping. I guess we could simply use STACK_NAME instead, which technically isn't valid in a hostname?

aermakov-zalando: I guess we could also use `$(STACK_NAME)`, which Kubernetes [supports](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config) for the pod command/arguments?

jonathanbeber: +1 for `$(STACK_NAME)`.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -96,6 +96,12 @@ type EmbeddedObjectMeta struct {
 	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,12,rep,name=annotations"`
 }
 
+type IngressOrRouteGroupSpec interface {
```
jonathanbeber, line 99: nit: I'd keep the interface close to where it's being used, I don't think any other abstraction but the `core` package will make use of that.

pkg/apis/zalando.org/v1/types.go
aryszka: how about just using a simple `?` as a placeholder? We don't need other templating in the overrides, or do we?

Another question: why 'must'? What happens if somebody just uses a full hostname without replacement? I don't really anticipate such a use case, but I also don't see always enforcing the stackname replacement in every case.

docs/stackset_crd.yaml
```diff
@@ -99,6 +99,38 @@ spec:
...
+                    description: Settings for the per-stack ingresses
+                    properties:
+                      enabled:
+                        description: Whether to enable per-stack ingresses or routegroups.
+                          Defaults to enabled if unset.
+                        type: boolean
+                      hosts:
+                        description: Hostnames to use for the per-stack ingresses
+                          (or route groups). These must contain the special $stack_name$
+                          token, which will replaced with the stack's name accordingly.
```
aryszka, line 112: s/which will replaced/which will be replaced/

docs/stackset_crd.yaml
```diff
@@ -238,6 +270,38 @@ spec:
...
+                    description: Settings for the per-stack route groups
+                    properties:
+                      enabled:
+                        description: Whether to enable per-stack ingresses or routegroups.
+                          Defaults to enabled if unset.
+                        type: boolean
+                      hosts:
+                        description: Hostnames to use for the per-stack ingresses
+                          (or route groups). These must contain the special $stack_name$
+                          token, which will replaced with the stack's name accordingly.
```
aryszka, line 283: s/which will replaced/which will be replaced/

pkg/apis/zalando.org/v1/types.go
aermakov-zalando: `$(STACK_NAME)` seems a bit more readable and explicit to me, `?` is kinda cryptic.

The 'must' is there because each stack's ingress needs to have a different hostname. Since the error reporting on external-dns/ingress-controller side is basically non-existent and the overlaps are impossible to debug, it's easier to just require this as early as possible.

pkg/core/stack_resources.go
```diff
@@ -299,26 +300,53 @@ func (sc *StackContainer) GenerateService() (*v1.Service, error) {
...
 
-	if len(clusterDomains) == 0 {
+	return result.List(), nil
+}
+
+func stackIngressRouteGroupAnnotations(spec zv1.IngressOrRouteGroupSpec) map[string]string {
+	if spec.GetOverrides() == nil || len(spec.GetOverrides().EmbeddedObjectMetaWithAnnotations.Annotations) == 0 {
+		return spec.GetAnnotations()
+	}
+	return spec.GetOverrides().EmbeddedObjectMetaWithAnnotations.Annotations
```
aryszka, line 338: Have you considered merging the annotations? I am just wondering which behavior the users would expect more often, this or overriding only specific annotations.

docs/stackset_crd.yaml
```diff
@@ -99,6 +99,38 @@ spec:
                     type: object
                   path:
                     type: string
+                  stackOverrides:
```
mikkeloscar, line 102: Would it make sense to model this as fields on the stack resource i.e. under the `stackTemplate`? It could still work the same way where what you define will overwrite and if you don't define anything then it defaults to the current behavior?

pkg/core/stack_resources.go
aermakov-zalando: I thought about this, but then there wouldn't be any way to unset an annotation. Not sure how much of a problem this would be.

docs/stackset_crd.yaml
aermakov-zalando: I started like that initially, but then I couldn't come up with a name for it that didn't suck. Any ideas? :)

pkg/core/stack_resources.go
aryszka: fair. Let's go with this approach :+1:

aryszka:
:+1:

docs/stackset_crd.yaml
jonathanbeber: maybe the same names, as overriding it?

aermakov-zalando: Yeah, but `stackIngressOverrides`/`stackRouteGroupOverrides` is rather annoying to spell out, and you'd also have to handle the conflicting cases (`ingress` + `stackRouteGroupOverrides`).

pkg/core/stack_resources.go
aermakov-zalando: Changed to merge, I guess it does make a bit more sense.

docs/stackset_crd.yaml
mikkeloscar: Why not simply call it `ingress` and `routegroup` and whatever you define there will be overwriting what would normally be created?

jonathanbeber: yes, that what I meant, the exact same names as you have inside the stackset. It just would be an extra boolean field `enabled`

pkg/core/stack_resources.go
jonathanbeber: how do we disable annotations like SSL policies in this case? Set it to empty?

docs/stackset_crd.yaml
aermakov-zalando: Sorry, I think I've misread the original proposal. Definitely makes more sense this way, will update it!

pkg/core/stack_resources.go
aermakov-zalando: I've rolled this change back for now, not sure if it makes sense now that they're moved out.

docs/stackset_crd.yaml
aermakov-zalando: Done, PTAL.

jonathanbeber:
:+1:

aermakov-zalando:
👎

pkg/apis/zalando.org/v1/types.go
```diff
@@ -96,17 +96,47 @@ type EmbeddedObjectMeta struct {
 	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,12,rep,name=annotations"`
 }
 
+// +k8s:deepcopy-gen=true
+type StackIngressRouteGroupOverrides struct {
+	EmbeddedObjectMetaWithAnnotations `json:"metadata,omitempty"`
+
+	// Whether to enable per-stack ingresses or routegroups. Defaults to enabled if unset.
+	// +optional
+	Enabled *bool `json:"enabled"`
```
mikkeloscar, line 105: nit: We could set the default to true in the schema.

```
// +kubebuilder:default:=true
```

aermakov-zalando: I'm not sure if this actually works at all (I think I've had some issues when I've tried it in a different CRD), and I'm not sure if we even want to do this in the first place since this means the stacksets and stacks would _always_ have `.ingress { enabled: true }` + `.routegroup { enabled: true }` even if the users aren't using those.

mikkeloscar: > this means the stacksets and stacks would always have .ingress { enabled: true } + .routegroup { enabled: true } even if the users aren't using those.

Good point, then it's maybe more confusing than helpful.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #330, opened 2021-06-10
Fix the overrides on 1.19, add E2E

aermakov-zalando:
Follow-up to #329, add E2E tests.

This also fixes the issue with the overrides not working on 1.19 unless `enabled` is set in the template.


aermakov-zalando:
👍

aryszka:
:+1:

Issue #340, opened 2021-07-12
Routegroup Algorithm Support


herojan:
Right now, I don't believe there is a way to specify the load balancer algorithm to be used for the generated default backends in a route group.

i.e. https://github.com/zalando-incubator/stackset-controller/blob/master/pkg/core/stackset.go#L168 uses the default algorithm.

I would like to be able to say for the default backend to use consistenthash load balancer, for example.

herojan:
@mikkeloscar @szuecs Does this feature request make sense? I made a pr for it anyway which I think works.

mikkeloscar:
I think it does make sense. Seems like we forgot/didn't consider this use case when we first added support. I'll take a closer look at your PR a bit later today.

PR #341, opened 2021-07-12
Add support for default load balancing algorithm

herojan:
Closes https://github.com/zalando-incubator/stackset-controller/issues/340

I'm happy to modify naming or docs if needed


herojan:
Actually I see I need to pass the algorithm down to the stacks too, will add that soon.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -156,9 +156,9 @@ type RouteGroupSpec struct {
 	AdditionalBackends []rg.RouteGroupBackend `json:"additionalBackends,omitempty"`
 	// Routes is the list of routes to be applied to the routegroup.
 	// +kubebuilder:validation:MinItems=1
-	Routes           []rg.RouteGroupRouteSpec `json:"routes"`
-	BackendPort      int                      `json:"backendPort"`
-	BackendAlgorithm string                   `json:"algorithm,omitempty"`
+	Routes      []rg.RouteGroupRouteSpec `json:"routes"`
+	BackendPort int                      `json:"backendPort"`
+	LBAlgorithm string                   `json:"lbAlgorithm,omitempty"`
```
mikkeloscar, line 161: Here we should use the type `BackendAlgorithmType` : https://github.com/szuecs/routegroup-client/blob/1b69df565b429b5aa497a3ff11904ff345abb2a1/apis/zalando.org/v1/types.go#L66 such that the generated CRD will have the right ENUM spec.

Additionally it should be marked optional: `// +optional` and we should add a comment describing that it applies to the backends generated for the stacks.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -156,9 +156,9 @@ type RouteGroupSpec struct {
 	AdditionalBackends []rg.RouteGroupBackend `json:"additionalBackends,omitempty"`
 	// Routes is the list of routes to be applied to the routegroup.
 	// +kubebuilder:validation:MinItems=1
-	Routes           []rg.RouteGroupRouteSpec `json:"routes"`
-	BackendPort      int                      `json:"backendPort"`
-	BackendAlgorithm string                   `json:"algorithm,omitempty"`
```
mikkeloscar: Hmm, seems like we already support it indirectly via this? Maybe we just need to change the json field name to `backendAlgorithm` and we avoid changing the spec to `lbAlgorithm`.

herojan: BackendAlgorithm was a field I added in my first commit, I changed it to lbAlgorithm in a second commit since it seemed clearer

mikkeloscar: 🤦‍♂️ was looking at the wrong diff

docs/stackset_crd.yaml
```diff
@@ -152,6 +152,9 @@ spec:
                     type: array
                   backendPort:
                     type: integer
+                  lbAlgorithm:
+                    description: The load balancing algorithm for the generated default backend to use.
+                    type: string
```
mikkeloscar, line 157: could you run `make` and check in the `stackset_crd.yaml` changes as well? It should generate an `enum` for the spec with the set of available algorithms.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -158,6 +158,8 @@ type RouteGroupSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	Routes      []rg.RouteGroupRouteSpec `json:"routes"`
 	BackendPort int                      `json:"backendPort"`
+	// +optional
+	LBAlgorithm rg.BackendAlgorithmType `json:"lbAlgorithm,omitempty"`
```
mikkeloscar, line 162: nit: As I mentioned in a previous comment it would be good to have a description comment on this field, similar to what was there before.

```
The load balancing algorithm used for the generated per stack backends.
```

herojan: Ah missed that somehow, added

jonathanbeber:
Thanks for this one, @herojan, could you please run a `go mod tidy`?

herojan:
@jonathanbeber Done

herojan:
I verified on a k8s cluster that:

1. Wrong values for lbAlgorithm are rejected because it's an enum
1. Updating a stackset routegroup with lbAlgorithm makes the main routegroup defaultBackend have algorithm, per stack routegroup doesn't
1. When new stack rolls out the per-stack ingress has the algorithm set

docs/stackset_crd.yaml
```diff
@@ -119,11 +119,17 @@ spec:
...
+                          - roundRobin
+                          - random
+                          - consistentHash
+                          - powerOfRandomNChoices
                           type: string
                         endpoints:
                           description: Endpoints is required for Type lb
                           items:
                             type: string
+                          minItems: 1
```
aryszka, line 132: I am not sure if need the `minItems` here. At the time we specify the stackset, we don't know about the endpoints. The stackset will generate a backend of type service (the algorithm field is valid for service backends, too). Then skipper based on this service backend, fetches the current endoints, and generates an LB backend with the current endpoints.

mikkeloscar: Note this is auto generated from the RouteGroup specification. It was added in this PR because of an update to the RouteGroup client: `github.com/szuecs/routegroup-client`

This is under the `additionalBackends` sections, so what is NOT generated by the stackset, but rather custom backends defined by the users. In this case it IMO makes sense to enforce the minimum if users specify an `lb` type backend.

aryszka: > This is under the additionalBackends sections

I see, thanks for clarifying. In this case it makes perfect sense.

aryszka:
:+1:

mikkeloscar:
:+1:

PR #343, opened 2021-07-31
Support RouteGroup autoscaler

mikkeloscar:
Adds support for scaling based on RPS by referencing a `RouteGroup` similar to what is possible for `Ingress`.

Additionally it uses the dedicated backend label for ingress and routegroup HPA config as introduced with https://github.com/zalando-incubator/kube-metrics-adapter/pull/350

Depends on:
* https://github.com/zalando-incubator/kube-metrics-adapter/pull/349
* https://github.com/zalando-incubator/kube-metrics-adapter/pull/350

PR #354, opened 2021-09-27
Revert dedicated backend label in ingress HPA

mikkeloscar:
Reverting part of #343 where label based backend identification was introduced. Select backend based on label does not work in the kube-metrics-adapter as it was not fully implemented in https://github.com/zalando-incubator/kube-metrics-adapter/pull/350

Without this multiple stacks can cause the RPS based metrics to flap because the kube-metrics-adapter randomly returns one of the stacks without considering the backend label value.

This does not fix RouteGroup as that depends on the broken label based identifier.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #365, opened 2021-11-03
Add the type match label

jonathanbeber:
When using the `autoscaler` we are still letting `kube-metrics-adapter`
to rely in the metric name. This commit adds the `type` label to the
ZMON and SQS queue len metrics. It also sorts the metrics before
generating them, based on different criteria.

Previously, sort was based just on the HPA metric type. Now sort is done
by `StackSet`'s `autoscaler` metric type and once it conflicts
`Average`, `AverageUtilization`, etc. It guarantees we support multiple
metrics of the same type and do not regenerate the HPA if Kubernetes
sends the metrics in a different order.

### DIFF
--- a/pkg/core/autoscaler.go
+++ b/pkg/core/autoscaler.go
@@ -17,21 +17,24 @@ import (
 )
 
 const (
-	requestsPerSecondName        = "requests-per-second"
-	metricConfigJSONPath         = "metric-config.pods.%s.json-path/path"
-	metricConfigJSONKey          = "metric-config.pods.%s.json-path/json-key"
-	metricConfigJSONPort         = "metric-config.pods.%s.json-path/port"
-	sqsQueueLengthTag            = "sqs-queue-length"
-	zmonCheckMetricName          = "zmon-check"
-	zmonCheckCheckIDTag          = "check-id"
-	zmonCheckAggregatorsTag      = "aggregators"
-	zmonCheckDurationTag         = "duration"
-	zmonCheckStackTag            = "stack"
-	zmonCheckKeyAnnotation       = "metric-config.external.zmon-check.zmon/key"
-	zmonCheckTagAnnotationPrefix = "metric-config.external.zmon-check.zmon/tag-"
-	sqsQueueNameTag              = "queue-name"
-	sqsQueueRegionTag            = "region"
-	scalingScheduleAPIVersion    = "zalando.org/v1"
+	metricsTypeLabel                   = "type"
+	requestsPerSecondName              = "requests-per-second"
+	metricConfigJSONPath               = "metric-config.pods.%s.json-path/path"
+	metricConfigJSONKey                = "metric-config.pods.%s.json-path/json-key"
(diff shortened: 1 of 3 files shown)


jonathanbeber:
@mikkeloscar I didn't investigate too deep, but from what I understood we don't have any other metric in the `autoscaler` type that could use the type too.

Also, I'm not sure if we would have any advantage changing the metric name field, since we don't have a better information but use the type as we have been doing.

pkg/core/autoscaler.go
```diff
@@ -313,6 +318,7 @@ func zmonMetric(metrics zv1.AutoscalerMetrics, stackName, namespace string) (*au
 				Name: zmonCheckMetricName,
 				Selector: &metav1.LabelSelector{
 					MatchLabels: map[string]string{
+						metricsTypeLabel:     zmonCheckMetricName,
```
mikkeloscar, line 321: Here the type name should just be `zmon` (https://github.com/zalando-incubator/kube-metrics-adapter#zmon-collector)

We probably also want to generate a unique `Name` per metric so if you have multiple SQS or ZMON metrics defined they will have a different name each.

jonathanbeber: Good catch, we implemented it now.

pkg/core/autoscaler.go
```diff
@@ -310,9 +376,10 @@ func zmonMetric(metrics zv1.AutoscalerMetrics, stackName, namespace string) (*au
 		Type: autoscaling.ExternalMetricSourceType,
 		External: &autoscaling.ExternalMetricSource{
 			Metric: autoscaling.MetricIdentifier{
-				Name: zmonCheckMetricName,
+				Name: fmt.Sprintf("%s-%d", zmonCheckMetricName, position),
```
mikkeloscar, line 379: The annotation we set below on line 403 refers to the metric name. So the annotation key must also be dynamic to work with this dynamic name.

mikkeloscar: This has been fixed!

pkg/core/autoscaler.go
```diff
@@ -42,43 +45,106 @@ var (
 	errMissingClusterScalingScheduleName       = errors.New("missing ClusterScalingSchedule metric object name")
 )
 
-type MetricsList []autoscaling.MetricSpec
+type autoscalerMetricsList []zv1.AutoscalerMetrics
 
-func (l MetricsList) Len() int {
+func (l autoscalerMetricsList) Len() int {
```
gargravarr, line 50: If the type is private, I think this method should be private too.

pkg/core/autoscaler.go
```diff
@@ -42,43 +45,106 @@ var (
...
 	return len(l)
 }
 
-func (l MetricsList) Swap(i, j int) {
+func (l autoscalerMetricsList) Swap(i, j int) {
```
gargravarr, line 54: If the type is private, I think this method should be private too.

gargravarr:
:+1:

mikkeloscar:
:+1:

PR #385, opened 2022-02-17
Sort stack cleanup candidates by NoTrafficSince

zaklawrencea:
Currently old Stacks are cleaned up based on the CreationTimestamp. However, older stacks may have received traffic more recently compared to a newer stack. This change evaluates the stacks to be cleaned based on noTrafficSince instead of CreationTimestamp.

Signed-off-by: zlawrence <zak.lawrence@zalando.de>

### DIFF
--- a/pkg/core/stackset.go
+++ b/pkg/core/stackset.go
@@ -116,10 +116,9 @@ func (ssc *StackSetContainer) MarkExpiredStacks() {
 		return
 	}
 
-	// sort candidates by oldest
+	// sort condidates by when they last had traffic.
 	sort.Slice(gcCandidates, func(i, j int) bool {
-		// TODO: maybe we use noTrafficSince instead of CreationTimeStamp to decide oldest
-		return gcCandidates[i].Stack.CreationTimestamp.Time.Before(gcCandidates[j].Stack.CreationTimestamp.Time)
+		return gcCandidates[i].Stack.Status.NoTrafficSince.Time.Before(gcCandidates[j].Stack.Status.NoTrafficSince.Time)
 	})
 
 	excessStacks := len(gcCandidates) - historyLimit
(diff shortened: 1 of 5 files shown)


pkg/core/stackset.go
```diff
@@ -116,10 +116,9 @@ func (ssc *StackSetContainer) MarkExpiredStacks() {
 		return
 	}
 
-	// sort candidates by oldest
+	// sort condidates by when they last had traffic.
 	sort.Slice(gcCandidates, func(i, j int) bool {
-		// TODO: maybe we use noTrafficSince instead of CreationTimeStamp to decide oldest
-		return gcCandidates[i].Stack.CreationTimestamp.Time.Before(gcCandidates[j].Stack.CreationTimestamp.Time)
+		return gcCandidates[i].Stack.Status.NoTrafficSince.Time.Before(gcCandidates[j].Stack.Status.NoTrafficSince.Time)
```
szuecs, line 121: this panics in test.

I guess `Status.NoTrafficSince` is not set, yet or not anymore.

pkg/core/stackset_test.go
```diff
@@ -31,22 +31,22 @@ func TestExpiredStacks(t *testing.T) {
...
 	}{
 		{
-			name:    "test GC oldest stack",
+			name:    "test GC stack last received traffic",
 			limit:   1,
 			ingress: true,
 			stacks: []*StackContainer{
 				testStack("stack1").createdAt(now.Add(-1 * time.Hour)).noTrafficSince(now.Add(-1 * time.Hour)).stack(),
-				testStack("stack2").createdAt(now.Add(-2 * time.Hour)).noTrafficSince(now.Add(-1 * time.Hour)).stack(),
+				testStack("stack2").createdAt(now.Add(-2 * time.Hour)).noTrafficSince(now.Add(-2 * time.Hour)).stack(),
```
mikkeloscar, line 39: We should also have a test case where the oldest stack (creation time) has a newer `noTrafficSince` time to verify that this case is now also working. With the old logic these test cases would still pass :)

zaklawrencea: Good spot, added test case.

pkg/core/stackset.go
zaklawrencea: Thanks, this is resolved now.

mikkeloscar:
:+1:

zaklawrencea:
👍

PR #396, opened 2022-04-12
Bug/reconcile ignores rg annotations

gusampaio:


mikkeloscar:
Thanks for adding a fix. I made a PR just with e2e tests #397 to show it currently doesn't work and verify if your change does.

PR #397, opened 2022-04-14
Test for annotation management on sub resources

mikkeloscar:
Copy of #396 for running e2e (only possible on local branches)

This also adds e2e tests and fixes one unit test case which was not covered correctly with the change.

---


katyanna:
:+1:

mikkeloscar:
:+1:

PR #398, opened 2022-04-14
Bug/reconcile ignores rg annotations

mikkeloscar:
Copy of #396 for running e2e (only possible on local branches)

---


mikkeloscar:
:+1:

controller/stackset.go
```diff
@@ -687,14 +687,16 @@ func (c *StackSetController) AddUpdateStackSetIngress(ctx context.Context, stack
...
+	lastUpdateValue, existingHaveUpdateTimeStamp := existing.Annotations[ControllerLastUpdatedAnnotationKey]
 	if existingHaveUpdateTimeStamp {
 		delete(existing.Annotations, ControllerLastUpdatedAnnotationKey)
 	}
 
 	// Check if we need to update the Ingress
 	if existingHaveUpdateTimeStamp && equality.Semantic.DeepDerivative(ingress.Spec, existing.Spec) &&
 		equality.Semantic.DeepEqual(ingress.Annotations, existing.Annotations) {
+		// add the annotation back after comparing
+		existing.Annotations[ControllerLastUpdatedAnnotationKey] = lastUpdateValue
```
szuecs, line 699: maybe the logic could be simplified:
```
	if existingHaveUpdateTimeStamp && equality.Semantic.DeepDerivative(ingress.Spec, existing.Spec) &&
		equality.Semantic.DeepEqual(ingress.Annotations, existing.Annotations) {
		// add the annotation back after comparing
		existing.Annotations[ControllerLastUpdatedAnnotationKey] = lastUpdateValue
		return existing, nil
	} else if existingHaveUpdateTimeStamp {
		delete(existing.Annotations, ControllerLastUpdatedAnnotationKey)
	}

mikkeloscar: The intention of the logic is to compare the annotations of the existing and of the newly generated resource. Since the newly generated resource doesn't have the `ControllerLastUpdatedAnnotationKey` we delete it to only compare all the rest. If they are equal we don't need the newly generated resource and therefore return the existing one, but we want to return it with the `ControllerLastUpdatedAnnotationKey` key to ensure we compare the real resource state at a later point.

That's why the logic was done like this.

I'll think about doing this in a nicer way so it's maybe more clear.

szuecs: From my side it's fine, but if I see 2x `if $foo` within 5 lines, I think we can put them together

szuecs:
:+1:

PR #411, opened 2022-06-27
Build multi-arch image

mikkeloscar:
Build multi-arch image along with the open source single arch image. Use multi-arch images in e2e tests.


mikkeloscar:
:+1:

PR #412, opened 2022-06-28
Use cdp-runtime for the build prod step

mikkeloscar:
The production build from #411 failed because the CDP step was using the legacy golang overlay instead of the new cdp-runtime.

This fixes it by using the cdp-runtime which will also introduce the latest Go version :)


linki:
👍

PR #424, opened 2022-08-01
Fix label update on stackset ingress

demonCoder95:
An [issue](https://github.bus.zalan.do/zooport/issues/issues/2677) was raised where the change of labels on the "parent" `Stackset` ingress was not reconciled with the existing object.

This fixes it by:
1. Updating the condition for reconciliation (to not ignore `ingress` update if a label has changed)
2. Updating labels on the "updating" `ingress` so they can be patched to the existing object in the cluster
3. Adding a unit test 

Update: The same was fixed for the stackset `RouteGroup` as well.

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -694,7 +694,8 @@ func (c *StackSetController) AddUpdateStackSetIngress(ctx context.Context, stack
 
 	// Check if we need to update the Ingress
 	if existingHaveUpdateTimeStamp && equality.Semantic.DeepDerivative(ingress.Spec, existing.Spec) &&
-		equality.Semantic.DeepEqual(ingress.Annotations, existing.Annotations) {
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I think we have the same bug for routegroup, could we fix it there at the same time?

demonCoder95:
Sure, good idea.

demonCoder95:
👍

mikkeloscar:
@demonCoder95 Can we also get unit tests for the routegroup part :) then it lgtm.

demonCoder95:
> @demonCoder95 Can we also get unit tests for the routegroup part :) then it lgtm.

Oh, I added the unit test for the routegroup, no? 🤔

mikkeloscar:
:+1:

mikkeloscar:
@demonCoder95 sorry. I missed it in the diff :(

PR #432, opened 2022-09-13
Address staticcheck warnings

mikkeloscar:
Address staticcheck issues

```
staticcheck ./...
cmd/e2e/broken_stack_test.go:9:2: package "k8s.io/api/core/v1" is being imported more than once (ST1019)
	cmd/e2e/broken_stack_test.go:10:2: other import of "k8s.io/api/core/v1"
cmd/e2e/test_utils.go:18:2: package "k8s.io/api/core/v1" is being imported more than once (ST1019)
	cmd/e2e/test_utils.go:19:2: other import of "k8s.io/api/core/v1"
cmd/e2e/test_utils.go:369:2: should use 'return err == nil' instead of 'if err != nil { return false }; return true' (S1008)
cmd/e2e/test_utils.go:378:2: should use 'return err == nil' instead of 'if err != nil { return false }; return true' (S1008)
cmd/e2e/test_utils.go:469:32: type assertion to the same type: err already has type error (S1040)
cmd/e2e/test_utils.go:469:46: type assertion to the same type: err already has type error (S1040)
cmd/e2e/test_utils.go:483:32: type assertion to the same type: err already has type error (S1040)
cmd/e2e/test_utils.go:483:46: type assertion to the same type: err already has type error (S1040)
cmd/traffic/main.go:75:2: should merge variable declaration with assignment on next line (S1021)
controller/stackset.go:23:2: package "k8s.io/api/core/v1" is being imported more than once (ST1019)
	controller/stackset.go:24:2: other import of "k8s.io/api/core/v1"
controller/stackset.go:43:2: const stackTrafficWeightsAnnotationKey is unused (U1000)
controller/stackset.go:44:2: const ingressAuthorativeAnnotationKey is unused (U1000)
pkg/core/stack_resources.go:76:6: func embeddedToObjectMeta is unused (U1000)
```

Also updates the golangci configuration so it will find such issues in the future.

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -20,7 +20,6 @@ import (
 	"github.com/zalando-incubator/stackset-controller/pkg/core"
 	"github.com/zalando-incubator/stackset-controller/pkg/recorder"
 	"golang.org/x/sync/errgroup"
-	apiv1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	networking "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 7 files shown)


controller/stackset.go
```diff
@@ -162,6 +158,7 @@ func (c *StackSetController) Run(ctx context.Context) {
 			var reconcileGroup errgroup.Group
 			for stackset, container := range stackContainers {
 				container := container
+				stackset := stackset
```
mikkeloscar, line 161: Missing this could have been contributing to the flaky e2e tests if the reconcileGroup loop below end up skipping some stacksets because the variable is overwritten in the loop 🤔

mikkeloscar:
:+1:

katyanna:
👍

PR #434, opened 2022-09-15
Limit how many stacksets are reconciled in parallel

mikkeloscar:
While debugging flaky stackset e2e tests in `kubernetes-on-aws` I observed many logs in the stackset-controller with timeouts while POSTing changes to the apiserver e.g. creating/updating resources as part of the reconciliation loop.

We currently try to reconcile ALL stacksets in parallel which which is maybe too much load on either the single kube client we use or on the server. In either case we don't need to reconcile everything in parallel but can limit this to e.g. 10 in parallel at a time to reduce the load.

This PR introduces a flag `reconcile-workers` which controls how many stacksets can be reconciled in parallel. The default is set to 10 which seems like a reasonable number. It can be disabled by setting the flag to `-1` meaning unlimited.

### DIFF
--- a/cmd/stackset-controller/main.go
+++ b/cmd/stackset-controller/main.go
@@ -26,6 +26,7 @@ const (
 	defaultIngressSourceSwitchTTL = "5m"
 	defaultMetricsAddress         = ":7979"
 	defaultClientGOTimeout        = 30 * time.Second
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

demonCoder95:
Let's hope this works 😁  🤞

PR #441, opened 2022-10-25
autoscaler: ensure HPA min replicas is not below Stack replicas

AlexanderYastrebov:
Users may scale stack manually via `kubectl scale stack` which updates Stack `spec.replicas` but it would not work when HPA is present and has lower minReplicas value.

This change ensures that Stack's HPA min replicas is not below Stack `spec.replicas` (if present).

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

### DIFF
--- a/pkg/client/clientset/versioned/fake/register.go
+++ b/pkg/client/clientset/versioned/fake/register.go
@@ -37,14 +37,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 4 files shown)


pkg/client/clientset/versioned/fake/register.go
```diff
@@ -37,14 +37,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//   )
+//	import (
```
AlexanderYastrebov, line 40: This chnage is due to go1.19 formatting chnages

szuecs:
:+1:

PR #456, opened 2023-01-17
Allow traffic switch before stack is 100% Ready

katyanna:
Given the need for switching traffic to a stack that isn't 100% ready yet,
the `minReadyPercent` StackSet field allows setting a minimum percentage
of ready replicas required to enable the traffic switch.

`minReadyPercent` is optional and expects values from 1 to 100. In case of
value equal to or under 0 or over 100, it's normalised to 100, which keeps
the original behaviour.

### DIFF
--- a/docs/stackset_crd.yaml
+++ b/docs/stackset_crd.yaml
@@ -103,6 +103,10 @@ spec:
                 - backendPort
                 - hosts
                 type: object
+              minReadyPercent:
+                description: minReadyPercent sets the minimum percentage of Pods expected
+                  to be Ready to consider a Stack for traffic switch
+                type: integer
               routegroup:
                 description: RouteGroup is an alternative to ingress allowing more
                   advanced routing configuration while still maintaining the ability
--- a/pkg/core/test_helpers.go
+++ b/pkg/core/test_helpers.go
@@ -45,10 +45,14 @@ func testStack(name string) *testStackFactory {
 }
 
 func (f *testStackFactory) ready(replicas int32) *testStackFactory {
+	return f.partiallyReady(replicas, replicas)
+}
+
+func (f *testStackFactory) partiallyReady(readyReplicas, replicas int32) *testStackFactory {
 	f.container.resourcesUpdated = true
 	f.container.deploymentReplicas = replicas
-	f.container.updatedReplicas = replicas
-	f.container.readyReplicas = replicas
+	f.container.updatedReplicas = readyReplicas
+	f.container.readyReplicas = readyReplicas
 	return f
 }
 
--- a/pkg/core/traffic.go
+++ b/pkg/core/traffic.go
@@ -21,6 +21,16 @@ func allZero(weights map[string]float64) bool {
 	return true
 }
 
+// normalizeMinReadyPercent normalizes minimum percentage of Ready pods.
+// If value is under or equal to 0, or over or equal to 100, set it to 1.0
+// If value is between 0-100, it's then set as decimal
+func normalizeMinReadyPercent(minReadyPercent int) float64 {
+	if minReadyPercent >= 100 || minReadyPercent <= 0 {
+		return 1.0
+	}
+	return float64(minReadyPercent) / 100
+}
+
 // normalizeWeights normalizes a map of backend weights.
 // If all weights are zero the total weight of 100 is distributed equally
 // between all backends.
@@ -155,9 +165,11 @@ func (ssc *StackSetContainer) ManageTraffic(currentTimestamp time.Time) error {
 		}
 	}
 
+	minReadyPercent := normalizeMinReadyPercent(ssc.StackSet.Spec.MinReadyPercent)
 	for stackName, stack := range stacks {
 		stack.desiredTrafficWeight = desiredWeights[stackName]
 		stack.actualTrafficWeight = actualWeights[stackName]
+		stack.minReadyPercent = minReadyPercent
 	}
 
 	// Run the traffic reconciler which will update the actual weights according to the desired weights. The resulting
--- a/pkg/core/traffic_test.go
+++ b/pkg/core/traffic_test.go
@@ -46,12 +46,108 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
 	}
 }
 
+func TestTrafficSwitchSimpleReadyByPercentage(t *testing.T) {
+	for _, tc := range []struct {
+		name               string
+		stack              *StackContainer
+		resourcesUpdated   bool
+		deploymentReplicas int32
+		minReadyPercent    int
+		updatedReplicas    int32
+		readyReplicas      int32
+		expectedError      string
+	}{
+		{
+			name:               "min percent reachead",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    85,
+			readyReplicas:      85,
+		},
+		{
+			name:               "min percent not reachead on updates",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    84,
+			readyReplicas:      85,
+			expectedError:      "stacks not ready: foo-v1",
+		},
+		{
+			name:               "min percent not reachead on readiness",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    85,
+			readyReplicas:      84,
+			expectedError:      "stacks not ready: foo-v1",
+		},
+		{
+			name:               "min percent reachead, decimal result",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    50,
+			updatedReplicas:    2,
+			readyReplicas:      2,
+		},
+		{
+			name:               "min percent under 0",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    -10, // normalized to 100
+			updatedReplicas:    3,
+			readyReplicas:      3,
+		},
+		{
+			name:               "min percent at 0",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    0, // normalized to 100
+			updatedReplicas:    3,
+			readyReplicas:      3,
+		},
+		{
+			name:               "min percent over 100",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    200, // normalized to 100
+			updatedReplicas:    3,
+			readyReplicas:      3,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c := StackSetContainer{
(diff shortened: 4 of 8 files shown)


pkg/core/test_helpers.go
```diff
@@ -52,6 +52,14 @@ func (f *testStackFactory) ready(replicas int32) *testStackFactory {
 	return f
```
linki, line 52: We can change the body of this function to `return partiallyReady(replicas, replicas)`.

pkg/core/test_helpers.go
```diff
@@ -52,6 +52,14 @@ func (f *testStackFactory) ready(replicas int32) *testStackFactory {
...
 }
 
+func (f *testStackFactory) partiallyReady(readyReplicas, replicas int32) *testStackFactory {
+	f.container.resourcesUpdated = true
+	f.container.deploymentReplicas = replicas
+	f.container.updatedReplicas = readyReplicas
```
linki, line 58: Let's use `replicas` here to really only have it non-ready (as opposed to also being non-updated).

pkg/core/traffic.go
```diff
@@ -158,6 +168,7 @@ func (ssc *StackSetContainer) ManageTraffic(currentTimestamp time.Time) error {
 	for stackName, stack := range stacks {
 		stack.desiredTrafficWeight = desiredWeights[stackName]
 		stack.actualTrafficWeight = actualWeights[stackName]
+		stack.minReadyPercent = normalizeMinReadyPercent(ssc.StackSet.Spec.MinReadyPercent)
```
linki, line 171: This value is not Stack-specific (like the two variables above). Every Stack in this loop will get the same `minReadyPercent` value. Instead of copying it down to the `Stack` (or `StackContainer`) couldn't we just read the value from the StackSet later on?

pkg/core/traffic_test.go
```diff
@@ -46,49 +46,158 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
...
+			readyReplicas:      3,
+			expectedError:      "stacks not ready: foo-v1",
+		},
+		{
+			name:               "min percent reachead",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    90,
+			readyReplicas:      90,
```
linki, line 84: I think we can drop this case (90/85). The 84/85 and 85/85 are probably enough.

docs/stackset_crd.yaml
```diff
@@ -103,6 +103,10 @@ spec:
                 - backendPort
                 - hosts
                 type: object
+              minReadyPercent:
```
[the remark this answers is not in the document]
katyanna: I just didn't come up with a good name here, do you have any propositions?

pkg/core/traffic_test.go
```diff
@@ -46,49 +46,158 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
...
+			readyReplicas:      84,
+			expectedError:      "stacks not ready: foo-v1",
+		},
+		{
+			name:               "min percent not reachead, decimal result",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    85,
+			updatedReplicas:    2,
+			readyReplicas:      3,
```
linki, line 75: Here we're mixing up `ready` and `updated` but our tests don't catch that. I think https://github.com/zalando-incubator/stackset-controller/pull/456/files#r1072337923 should fix that.

pkg/core/traffic_test.go
```diff
@@ -46,49 +46,158 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
...
+			expectedError:      "stacks not ready: foo-v1",
```
linki, line 76: Let's also add the following case so we cover an interesting rounding example:
```
deploymentReplicas: 3,
minReadyPercent:    50,
updatedReplicas:    3,
readyReplicas:      2,
```

Note, 2/3 fails when requiring 85%. But 2/3 is fine if we only require 50% (this test).

pkg/core/traffic.go
katyanna: `IsReady` is a `StackContainer` method, I didn't find a way of getting to `Stackset` through what we already have there. An option would be to pass this as an argument, which looked a bit messy to me.

pkg/core/traffic_test.go
```diff
@@ -46,49 +46,158 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
...
 		deploymentReplicas int32
+		minReadyPercent    int
 		updatedReplicas    int32
 		readyReplicas      int32
 	}{
 		{
 			name:               "deployment not updated yet",
 			resourcesUpdated:   false,
 			deploymentReplicas: 3,
+			minReadyPercent:    100,
```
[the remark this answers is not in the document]
katyanna: that is the first case in `TestTrafficSwitchSimpleReadyByPercentage`, no?

pkg/core/traffic_test.go
```diff
@@ -243,30 +477,32 @@ func TestTrafficSwitchSimple(t *testing.T) {
...
 			},
 			expectedActualWeights: map[string]float64{
-				"foo-v1": 0,
-				"foo-v2": 100,
+				"foo-v1": 70,
+				"foo-v2": 30,
 				"foo-v3": 0,
 			},
-			expectedError: "stacks not ready: foo-v1, foo-v3",
+			expectedError: "stacks not ready: foo-v2, foo-v3",
```
linki, line 497: Maybe it's the diff but this looks odd. It tests that traffic is successfully switched but also expects there to be an error about non-ready stacks. Let's double check why this succeeds (if it even does currently).

pkg/core/traffic.go
linki: But we could at least move it out of the loop? This way it's easier to see that `minPercentReady` is the same value for all Stacks.

pkg/core/traffic_test.go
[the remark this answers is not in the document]
linki: Yes, it would be good to have one test with an explicit `100` (and maybe also one with an explicit `0`). The other old tests could just leave out that field (like before).

pkg/core/traffic_test.go
katyanna: I think this test is incorrect and not needed, we should be good with testing the readiness for switching.

docs/stackset_crd.yaml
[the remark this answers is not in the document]
linki: Let's keep it this way.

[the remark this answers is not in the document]
linki: Let's also add it in the StackSet example in the README so we have it documented somewhere.

pkg/core/traffic_test.go
```diff
@@ -83,12 +170,21 @@ func TestTrafficSwitchSimpleNotReady(t *testing.T) {
 			updatedReplicas:    0,
 			readyReplicas:      0,
 		},
+		{
+			name:               "min percent not reachead",
+			resourcesUpdated:   true,
+			deploymentReplicas: 3,
+			minReadyPercent:    85,
+			updatedReplicas:    2,
+			readyReplicas:      3,
```
linki, line 179: We should think about how we want to treat `updated` and `ready`. Here I see 3 out of 3 are ready. Why is this an error?

pkg/core/traffic_test.go
```diff
@@ -205,7 +424,8 @@ func TestTrafficSwitchSimple(t *testing.T) {
 			expectedError: "stacks not ready: foo-v2",
 		},
 		{
-			name: "if there are no stacks with a desired traffic weight, and no stacks had traffic recently, send traffic to the earliest created stack",
+			name:            "if there are no stacks with a desired traffic weight, and no stacks had traffic recently, send traffic to the earliest created stack",
+			minReadyPercent: 100,
```
linki, line 428: Let's drop it for all the old tests (i.e. where only this line changed), except for the first one above.

pkg/core/traffic_test.go
```diff
@@ -46,12 +46,99 @@ func TestTrafficSwitchNoIngress(t *testing.T) {
...
+		updatedReplicas    int32
+		readyReplicas      int32
+		expectedError      string
+	}{
+		{
+			name:               "min percent reachead",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    100,
```
linki, line 65: Maybe that was a bad idea after all. It's probably easier to understand (and doesn't make much of a difference) if we treat `ready` and `updated` equally for that matter. Let's revert it and set `updatedReplicas` to the same value as `readyReplicas`. (This also means we should do the same in `partiallyReady()`)

pkg/core/traffic_test.go
katyanna: We're still comparing the `updatedReplicas`, let's drop it. https://github.com/zalando-incubator/stackset-controller/pull/456/files/b2aac68aa1f84137df5d8a599d444332c2666e1a#diff-717f128517472d3dc4091e67bfc0f4fe4c36e32096c5ef87d78f34cbc64d2399R147

katyanna: > Stacks are considered ready when all subresources have been updated, and we have enough replicas

`IsReady()` considers the update status, I think we should keep the behaviour. All subresources must be updates, but we apply the `minReadyPercent` to the `readyReplicas`.

pkg/core/types.go
```diff
@@ -139,8 +140,14 @@ func (sc *StackContainer) HasTraffic() bool {
...
-	return sc.resourcesUpdated && sc.deploymentReplicas > 0 && sc.deploymentReplicas == sc.updatedReplicas && sc.deploymentReplicas == sc.readyReplicas
+	// Calculate minmum required replicas for the Deployment to be considered ready
+	minRequiredReplicas := int32(math.Ceil(float64(sc.deploymentReplicas) * sc.minReadyPercent))
+
+	// Stacks are considered ready when all subresources have been updated
+	// and the minimum ready percentage is hit on and replicas
+	return (sc.resourcesUpdated && sc.deploymentReplicas > 0 &&
+		minRequiredReplicas <= sc.updatedReplicas &&
+		minRequiredReplicas <= sc.readyReplicas)
 }
```
myaser, line 151: Q: shoud we also check a percentage of `resourcesUpdated`

myaser:
👍🏻

pkg/core/traffic_test.go
```diff
@@ -83,12 +179,21 @@ func TestTrafficSwitchSimpleNotReady(t *testing.T) {
 			updatedReplicas:    0,
 			readyReplicas:      0,
 		},
+		{
+			name:               "min percent not reachead",
+			resourcesUpdated:   true,
+			deploymentReplicas: 100,
+			minReadyPercent:    85,
+			updatedReplicas:    100,
+			readyReplicas:      84,
```
linki, line 188: Let's also reduce `updatedReplicas` similar to the tests above?

linki:
👍

katyanna:
👍

PR #463, opened 2023-02-20
Update dependencies

mikkeloscar:
Update dependencies to address vulnerabilities and to use the v1.22.17 client-go version which we will soon run in all clusters @ Zalando.


mikkeloscar:
:+1:

katyanna:
👍

PR #464, opened 2023-02-28
Use opensource multi-arch skipper image on ghcr.io

linki:
Since https://github.com/zalando/skipper/issues/2114 there's a multi-arch image being pushed to ghcr.io which is public: https://github.com/zalando/skipper/pkgs/container/skipper

Let's use it here so that the stackset-controller e2e tests over at https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/delivery.yaml#L241 can run against an all arm64 cluster (https://github.com/zalando-incubator/kubernetes-on-aws/pull/5098)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ spec:
         spec:
           containers:
           - name: skipper
-            image: registry.opensource.zalan.do/teapot/skipper:latest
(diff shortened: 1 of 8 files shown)


szuecs:
👍

linki:
👍

PR #472, opened 2023-03-15
Clarify overscale calculation with prescaling a stack in docs

demonCoder95:
As discussed in [this](https://github.bus.zalan.do/zooport/issues/issues/3466) zooport issue, this clarifies the `overscaling` that might arise from using the `prescale` feature of the stackset controller with a high `minReplicas` count and slow traffic switching.

### DIFF
--- a/docs/howtos.md
+++ b/docs/howtos.md
@@ -347,6 +347,8 @@ This means that it might overscale for some minutes before the HPA kicks in and
 scales back down to the needed resources. Reliability is favoured over cost in
 the prescale logic.
 
+### Overscaling Example
+
 To explain further, the amount of this overscale is decided as follows. Imagine
 a case with `minReplicas: 20` and `maxReplicas: 40` and traffic is switched in
 steps as `1%`, `25%`, `50%` and `100%`. Additionally, imagine the existing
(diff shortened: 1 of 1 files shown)


docs/howtos.md
```diff
@@ -342,11 +342,32 @@ configure the time by setting the
...
+25% of 40 = 10 (20 from old stack + 20 from the new), so the new stack is
+scaled to 20 + 10 = 30 replicas before the switch.
+3. Before switching to `50%`, same follows, and the prescale is now done as
+50% of 50 = 25 (30 replicas of new stack + 20 replicas of the old), and so
+the new stack is to be scaled to 55 replicas, however, since `maxReplicas` is
+set to 40, stack size will be set to `40`.
+4. Similarly, when `100%` of the traffic is to be switched, the size of
+`maxReplicas` will be enforced.
+
+A good discussion regarding how this issue can arise is in [this issue](https://github.bus.zalan.do/zooport/issues/issues/3466).
```
mikkeloscar, line 369: These docs are public so it's not meaningful to link to this issue from here.

While this information is valid in the public repo, I think it's more valuable internally where we have CDP and the traffic switching steps etc. I.e. extend the cloud-docs with this example.

demonCoder95: Ah, good point. I'll remove the link from here and add in the cloud-docs. Thanks!

demonCoder95: Resolved in the new commit.

mikkeloscar:
:+1:

demonCoder95:
👍

PR #485, opened 2023-05-04
Update to autoscaling/v2

mikkeloscar:
With Kubernetes v1.23 the the autoscaling API used for Horizontal Pod Autoscalers was promoted from `autoscaling/v2beta2` to `autoscaling/v2`.

This updates the API used for getting and creating HPAs to the `autoscaling/v2` version. This effectively means that the lowest supported Kubernetes version is now Kubernetes v1.23.


VERSION
```diff
@@ -1 +1 @@
-v1.3
+v1.4
```
mikkeloscar, line 1: bumped the version to reflect that we make a breaking change (new minimum Kubernetes version).

mikkeloscar:
:+1:

katyanna:
👍

PR #495, opened 2023-06-07
Add support for traffic segments.

gargravarr:
This Pull Request (PR) adds support for traffic management through the use of
Skipper's [TrafficSegment](https://opensource.zalando.com/skipper/reference/predicates/#trafficsegment) predicate. By default this feature is not active for
StackSets, and can be activated/controlled in the following way:

* If a new StackSet includes the annotation `"stackset-controller.zalando.org/use-traffic-segments": "true"`, the controller will manage ingress/routegroup and traffic with a dedicated resource per stack (a Ingress/RouteGroup segment), using `TrafficSegment` to split traffic. This annotation is only active when you start the controller with both `--enable-traffic-segments` and `--annotated-traffic-segments` set.
* If someone (or some entity) updates an existing StackSet with the annotation
`"stackset-controller.zalando.org/use-traffic-segments": "true"`, the controller
will convert the StackSet to use `TrafficSegment` in the following way:
    1. Creates Ingress/RouteGroup segments for each Stack, with the segment
       corresponding to each Stack traffic weight.
    2. Deletes the central Ingress/RouteGroup, only after the controller created
       per Stack segments.
* Starting the controller with the option `--enable-traffic-segments` enables by default traffic segments on all StackSets. In this mode the controller will also convert StackSets using a central ingress resource as described above.
* Starting the controller with the option `--enable-traffic-segments` and `--annotated-traffic-segments` enables traffic segments only on StzasckSets with the annotation `"stackset-controller.zalando.org/use-traffic-segments": "true"`.

On StackSets with traffic segments, the controller switches traffic one segment
at a time, prioritizing first stacks with a higher increase in traffic share.

This PR also updates end 2 end tests to repeat the same tests also with StackSets with explicit support for Traffic Segments.

Finally, this PR removes deprecated `varcheck` and `deadcode` linters,
already covered by `unused`.


cmd/e2e/README.md
```diff
@@ -31,7 +31,9 @@ watch -n 10 "kubectl get -n foo stackset,stack,ing,ep,deployment"
 ```
 kubectl delete namespace foo; kubectl create namespace foo
 make
-./build/stackset-controller --apiserver=http://127.0.0.1:8001 --controller-id=foo
+./build/stackset-controller --apiserver=http://127.0.0.1:8001
+--enable-routegroup-support --controller-id=foo
+--cluster-domain=${CLUSTER_DOMAIN} --cluster-domain=${CLUSTER_DOMAIN_INTERNAL}
```
mikkeloscar, line 36: nit: there should be `\` for each line for the command to work as is.

controller/stackset.go
```diff
@@ -204,6 +223,11 @@ func (c *StackSetController) Run(ctx context.Context) {
 				continue
 			}
 
+			if c.injectSegmentAnnotation(ctx, &stackset) {
+				// Reconciler adds StackSet in the next loop
+				continue
+			}
```
mikkeloscar, line 229: The Run loop is just for maintaining the tracking of stacksets managed by the controller, injecting the annotation at this stage seems wrong, it should rather be done in `ReconcileStackSet()`

(Same for the logic above update/delete)

controller/stackset.go
```diff
@@ -234,6 +302,9 @@ func (c *StackSetController) collectResources(ctx context.Context) (map[types.UI
 		}
 
 		stacksetContainer := core.NewContainer(&stackset, reconciler, c.backendWeightsAnnotationKey, c.clusterDomains)
+		if stackset.Annotations[TrafficSegmentsAnnotationKey] == "true" {
+			stacksetContainer.EnableSegmentTraffic()
+		}
```
mikkeloscar, line 307: nit: I would simply move this to `core.NewContainer`

Alternative a method on the stacksetContainer that returns true if the annotation is set could be another way to handle it.

pkg/core/traffic.go
```diff
@@ -1,14 +1,204 @@
...
+			"limits cannot have a negative value (%f, %f)",
+			lower,
+			upper,
+		)
+
+	case upper < lower:
+		return fmt.Errorf(
+			"upper limit must be >= than lower limit (%f, %f)",
+			lower,
+			upper,
```
mikkeloscar, line 114: nit: The error message is a bit confusing, it says "upper limit" first and "lower limit" last but has "(lower,upper)". I would phrase it as "lower must be smaller or equal to upper".

pkg/core/traffic.go
```diff
@@ -203,6 +404,91 @@ func (ssc *StackSetContainer) ManageTraffic(currentTimestamp time.Time) error {
...
+//
+// Returns an ordered list of traffic segments, to ensure no gaps in traffic
+// assignment.
+func (ssc *StackSetContainer) ComputeTrafficSegments() (
+	[]TrafficSegment,
+	error,
+) {
+	var segments segmentList
+	weightDiffs := map[types.UID]float64{}
+	res := []TrafficSegment{}
```
mikkeloscar, line 418: nit: maybe call this changes or something like that, so it's clear it's a list of changed/new TrafficSegments

controller/stackset.go
gargravarr: When we finish migration we won't need this annotation, that's why I defined it outside.

pkg/core/traffic.go
gargravarr: You're right. Updated.

controller/stackset.go
gargravarr: Makes sense, and only need the coded snippet in 1 section. The only issue I've found is that the reconciler is not immediately able to perform the rest of its logic (like creating stacks) because the update with the segment annoation modified the StackSet resource. Tee controller outputs this error: 

```
ERRO[0111] Unable to create stack: Operation cannot be fulfilled on stacksets.zalando.org "pet-color": the object has been modified; please apply your changes to the latest version and try again  controller=stackset namespace=issue-3434 stackset=pet-color
ERRO[0111] Traffic reconciliation failed: no stacks to assign traffic to  controller=stackset namespace=issue-3434 stackset=pet-color
ERRO[0111] Unable to reconcile stackset resources: invalid ingress, no paths defined  controller=stackset namespace=issue-3434 stackset=pet-color
```

So if `ReconcileStackSet` needs to inject the segment annotation, it will inject and return immediately, and in the next loop will execute the remaining logic.

mikkeloscar: Interesting, we mitigate this issue like this in other cases: https://github.com/zalando-incubator/stackset-controller/blob/5e17048c8103e1e3c24d2c1cab12fe6e48d3c204/controller/stackset.go#L658-L669

Alternative would be to update the stacksetContainer with a new stackSet object which is returned when calling Update, that way the following calls don't have a conflict, that seems nicer when I think about it now :thinking:

controller/stackset.go
```diff
@@ -1057,6 +1354,23 @@ func (c *StackSetController) ReconcileStackSet(ctx context.Context, container *c
 		c.stacksetLogger(container).Errorf("Unable to reconcile stackset traffic: %v", err)
 	}
 
+	// This is to support both central and segment-based traffic.
+	if container.SupportsSegmentTraffic() {
+		// Update traffic segments. Proceed on errors.
+		err = c.ReconcileTrafficSegments(ctx, container)
```
mikkeloscar, line 1360: The way this is implemented is different from how it's generally done in stackset-controller, and it creates some situations where we need to update a resource multiple times, which could otherwise be avoided.

The general idea is that we model the desired state in the StackSetContainer object and after it has been modelled correctly the resources are created/updated based on that.

With TrafficSegment the logic is split in multiple stages. E.g. on line 1336 it calls `ReconcileStackResources` which will create an "empty/no-traffic" TrafficSegment Ingress/RouteGroup and then here it will update that same Ingress/RouteGroup after calculating the segment values. This leads to multiple updates of the resource, and from what I can tell it also means on the first iteration of a new stack the update of the TrafficSegment would fail because the resource is being edited.
Another way to do this would be to calculate the trafficsegment values per stack _before_ calling  `ReconcileStackResources` such that when we call `ReconcileStackResources` we have all the information needed and can generate the right resources with correct TrafficSegment values.

pkg/core/traffic.go
```diff
@@ -1,14 +1,204 @@
...
+		}
+
+		if res.IngressSegment != nil &&
+			(res.lowerLimit != lowerLimit || res.upperLimit != upperLimit) {
+
+			return nil, errors.New(
+				"mismatch in routegroup and ingress segment values",
+			)
+		}
+		res.lowerLimit, res.upperLimit = lowerLimit, upperLimit
```
mikkeloscar, line 84: Are we ensuring that there can't be both an `RouteGroupSegment` and `IngressSegment` on the same stack somewhere? 🤔

Otherwise this would overwrite what was set by the `IngressSegment` block above.

szuecs: there are people actually using both on purpose

gargravarr: @mikkeloscar line 84 only  "overwrites" what's in line 64 when the stack contains an ingress and a routegroup with the same limits (see the conditional block in lines 77-83).

The code intentionally allows having a RouteGroup and an Ingress at the same time.

controller/stackset.go
gargravarr: So in short, move Traffic Segment reconciliation before Stack resources reconciliation?

mikkeloscar: Yes, _and_ adjust it such that TrafficSegmentReconciliation only updates the local state in the StackSetContainer object, and `ReconcileStackResources` uses that state to generate the respective stack ingress/routegroup with respective segment predicate.
Right now `TrafficSegmentReconciliation` does both computing of the segment values and updates the ingress/routegroup resources.

pkg/core/traffic.go
mikkeloscar: Right, it's right on the lines above :facepalm:

controller/stackset.go
gargravarr: There is an additional problem which I am trying to solve here: with the current logic, the reconciler updates segments in the specific order given by `ComputeTrafficSegments`. This is necessary to avoid gaps in traffic. So we also need to add the logic to:
* Reconcile ingress resources with segments to update in the right order
* Reconcile remaining ingress resources

> from what I can tell it also means on the first iteration of a new stack the update of the TrafficSegment would fail because the resource is being edited.
I haven't observed this error in my tests, do you have an example? Remember that the first iteration of a new Stack is always one with no traffic (owl segment -> (0.0, 0.0)).

mikkeloscar: > I haven't observed this error in my tests, do you have an example? Remember that the first iteration of a new Stack is always one with no traffic (owl segment -> (0.0, 0.0)).

You're right it can't really happen in practice, so my comment is more about updating the ingress/routegroups in two places in the code.

> This is necessary to avoid gaps in traffic. So we also need to add the logic to:

Yes, and it's IMO fine to e.g. modify the order of stacks of the stacksetcontainer in place before the reconcile stack loop, or alternatively sort the stacks based on their ingress segment values right before applying the stacks in reconcile stack loop.

gargravarr: I updated the reconciler in the following way:
* `ReconcileTrafficSegment` updates the stack containers needing a segment update and return the ordered slice of segments to the main StackSet reconciler.
* The StackSet reconciler uses the returning slice of segments to reconcile first the Stacks needing segment updates, and only after reconciles the remaining Stacks.

The challenge here was in signaling to the Ingress/Routegroup reconciler that the ingress resource needs an update in the cluster. To signal this update, `ComputeTrafficSegment` sets a special annotation `to-update`: when `true`, the ingress/routegroup reconciler will update the resource with the updated traffic segment values.

mikkeloscar: The signal is just the state between the existing and the new, if it's not aligned it needs and update. I think it's cleaner to do it e.g. like this: https://github.com/zalando-incubator/stackset-controller/compare/ingress-version...ingress-version-compare

gargravarr: Yes, cleaner like that. I added your commit.

gargravarr: @mikkeloscar some end 2 end tests fail after your change:

```
06/12/2023 08:34:16 --- FAIL: TestStacksetCreateRouteGroup (34.43s)
06/12/2023 08:34:16 --- FAIL: TestStacksetCreateIngress (34.47s)
06/12/2023 08:34:36 --- FAIL: TestStacksetUpdateDeleteIngress (53.85s)
06/12/2023 08:34:36 --- FAIL: TestStacksetUpdateIngress (53.86s)
06/12/2023 08:34:36 --- FAIL: TestStacksetUpdateRouteGroup (54.56s)
06/12/2023 08:34:37 --- FAIL: TestStacksetUpdateDeleteRouteGroup (54.67s)
06/12/2023 08:38:47 FAIL
```

mikkeloscar: Yeah sorry, I didn't actually test it e2e, could have opened a PR against this one first, but just wanted to show the idea first.

gargravarr: Fixed, was just a missing configuration when generating an ingress/routegroup segment.

controller/stackset.go
```diff
@@ -226,6 +233,55 @@ func (c *StackSetController) Run(ctx context.Context) {
...
+			"Failed injecting segment annotation: %v",
+			err,
+		)
+		return false
+	}
+	c.recorder.Eventf(
+		stackSet,
+		v1.EventTypeNormal,
+		"UpdatedStackSet",
+		"Updated StackSet %s",
```
mikkeloscar, line 278: nit: Maybe describe that the annotation was injected in this event so it's more clear what the update was.

controller/stackset.go
```diff
@@ -964,10 +1060,123 @@ func (c *StackSetController) ReconcileStackSetIngressSources(
...
+		)
+		if err != nil {
+			return err
+		}
+
+		c.recorder.Eventf(
+			ssc.StackSet,
+			v1.EventTypeNormal,
+			"DeletedIngress",
+			"Deleted Ingress %s, StackSet conversion complete",
```
mikkeloscar, line 1113: Maybe we can also write a bit more context so users don't get confused what conversion this is about if they see the event.

controller/stackset.go
```diff
@@ -1012,12 +1221,6 @@ func (c *StackSetController) ReconcileStackSetDesiredTraffic(ctx context.Context
 }
 
 func (c *StackSetController) ReconcileStackResources(ctx context.Context, ssc *core.StackSetContainer, sc *core.StackContainer) error {
-	if c.configMapSupportEnabled {
-		err := c.ReconcileStackConfigMap(ctx, sc.Stack, sc.Resources.ConfigMaps, sc.UpdateObjectMeta)
-		if err != nil {
-			return c.errorEventf(sc.Stack, "FailedManageConfigMap", err)
-		}
-	}
```
mikkeloscar, line 1020: I think this was lost during rebasing?

controller/stackset.go
gargravarr: Updated with mention to conversion to traffic segments (updated also for the case of RouteGroups).

controller/stackset.go
gargravarr: Oops. But why did e2es passed? :slightly_frowning_face:

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #497, opened 2023-06-12
dependabot: group updates

szuecs:
dependabot: group updates
Go modules updates every week, because if we group we have only one PR

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,16 @@ updates:
   schedule:
     interval: monthly
     time: "07:00"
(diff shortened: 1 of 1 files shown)


mikkeloscar:
@szuecs Does this really work? In external-dns it's suggested that it's a special alpha feature enabled explicitly for the repo: https://github.com/kubernetes-sigs/external-dns/pull/3678#issue-1753065732

szuecs:
Oh I should read commit messages more closely 😀

katyanna:
Added on https://github.com/zalando-incubator/stackset-controller/pull/641.

PR #498, opened 2023-06-13
crd: reduce CRD size

AlexanderYastrebov:
# Why

When CRD is applied via `kubectl apply -f docs/stackset_crd.yaml` (aka client-side apply) kubectl stores
CRD content into the kubectl.kubernetes.io/last-applied-configuration annotation which has
size limit of 262144 bytes.
If the size of the annotation exceeds the limit, kubectl will fail with the following error:

	The CustomResourceDefinition "stacksets.zalando.org" is invalid: metadata.annotations: Too long: must have at most 262144 bytes

See https://github.com/kubernetes/kubectl/issues/712

The CRD contains a lot of descriptions for k8s.io types and controller-gen
does not allow to skip descriptions per field or per package,
see https://github.com/kubernetes-sigs/controller-tools/issues/441

# How

It removes descriptions starting at the deepest level of the yaml tree
until the size of the yaml converted to json is less than the maximum allowed annotation size.

### DIFF
--- a/docs/stack_crd.yaml
+++ b/docs/stack_crd.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -945,16 +944,6 @@ spec:
                                                   Gt, and Lt.
                                                 type: string
                                               values:
-                                                description: An array of string values.
-                                                  If the operator is In or NotIn,
-                                                  the values array must be non-empty.
-                                                  If the operator is Exists or DoesNotExist,
-                                                  the values array must be empty.
-                                                  If the operator is Gt or Lt, the
-                                                  values array must have a single
-                                                  element, which will be interpreted
-                                                  as an integer. This array is replaced
-                                                  during a strategic merge patch.
                                                 items:
                                                   type: string
                                                 type: array
@@ -983,16 +972,6 @@ spec:
                                                   Gt, and Lt.
                                                 type: string
                                               values:
-                                                description: An array of string values.
-                                                  If the operator is In or NotIn,
-                                                  the values array must be non-empty.
-                                                  If the operator is Exists or DoesNotExist,
-                                                  the values array must be empty.
-                                                  If the operator is Gt or Lt, the
-                                                  values array must have a single
-                                                  element, which will be interpreted
-                                                  as an integer. This array is replaced
-                                                  during a strategic merge patch.
                                                 items:
                                                   type: string
                                                 type: array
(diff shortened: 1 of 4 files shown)


docs/stack_crd.yaml
```diff
@@ -1,4 +1,3 @@
----
```
AlexanderYastrebov, line 1: Not sure if we need this and why controller-gen adds it, perhaps to enable concatenation via `cat foo.yaml bar.yaml > baz.yaml`

hack/crd/trim.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return descriptions[i].depth > descriptions[j].depth
+	})
+
+	for _, d := range descriptions {
+		if size <= maxAnnotationSize {
+			break
+		}
+		size -= len(d.value)
+		delete(d.property, "description")
```
AlexanderYastrebov, line 80: Here we can do more sophisticated logic, e.g. trim description to a single sentence like in https://github.com/zalando-incubator/kubernetes-on-aws/pull/5980

szuecs: interesting :)

szuecs:
Oh, oh:
```
When CRDs get too large, it becomes impossible to manage them with tools like Helm which store a compressed manifest in a Secret, which has a 1 MB limit.
```

RomanZavodskikh:
In theory, if the limit of `262144 bytes` would be increased, will we have a possibility to make descriptions full again?

docs/stack_crd.yaml
mikkeloscar: I think it's fine to now have it.

mikkeloscar:
:+1:

AlexanderYastrebov:
@RomanZavodskikh 
> In theory, if the limit of 262144 bytes would be increased, will we have a possibility to make descriptions full again?

Yes.

Another options are: use [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) that does not need annotation or create/update CRDs via API instead of using kubectl apply. Both require changes to https://github.com/zalando-incubator/cluster-lifecycle-manager/

Also StackSet embeds descriptions for for k8s.io types which may grow over time and CRD may theoretically hit the limit of k8s object size ([~1.5Mb](https://stackoverflow.com/questions/60468110/kubernetes-object-size-limitations)).

docs/stack_crd.yaml
AlexanderYastrebov: Did you mean _not_ to have it?

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #501, opened 2023-06-23
Add support for External RPS metrics on autoscaler

lucastt:
Adding support to the new feature from kube-metrics-adapter.


mikkeloscar:
@lucastt Can you also add documentation in this section similar to what we have for the other types? https://github.com/zalando-incubator/stackset-controller/blob/master/docs/howtos.md#specifying-horizontal-pod-autoscaler

pkg/apis/zalando.org/v1/types.go
```diff
@@ -280,6 +280,7 @@ const (
 	ZMONAutoscalerMetric         AutoscalerMetricType = "ZMON"
 	ClusterScalingScheduleMetric AutoscalerMetricType = "ClusterScalingSchedule"
 	ScalingScheduleMetric        AutoscalerMetricType = "ScalingSchedule"
+	ExternalRPSMetric            AutoscalerMetricType = "ExternalRPS"
```
lucastt, line 283:
```suggestion
	ExternalRPSMetric            AutoscalerMetricType = "requests-per-second"
```

mikkeloscar: Maybe we call it `RPS` or `RequestsPerSecond` such that the format is the same as we use for the other ones?

docs/howtos.md
```diff
@@ -250,6 +251,24 @@ autoscaler:
...
+  metrics:
+  - type: requests-per-second
+    average: 30
+    hostname: 'example.com;foo.bar.baz'
+    weight: 100
+```
+Note that `weight` field is not mandatory, in case it is not specified it will be assumed the value is 100%, and `hostname`
+accepts multiple hostnames using `;` as separator the result is the sum of average RPS of each hostname will be compared to
+the target specified by `average` field.
+
```
AlexanderYastrebov, line 271: Maybe we could make a bit clear how hostname (esp. when multiple are specified) and weight are related.
As far as I can tell from here https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/README.md#example-external-metric-1 this translates into annotations for kube-metrics-adapter

Also it looks like multiple hostnames should be separated by comma, not semicolon, shouldn't they?

docs/howtos.md
```diff
@@ -250,6 +251,24 @@ autoscaler:
...
+a external metric like `requests-per-second` might be pretty handy, see the example:
+
+```yaml
+autoscaler:
+  minReplicas: 1
+  maxReplicas: 3
+  metrics:
+  - type: requests-per-second
+    average: 30
+    hostname: 'example.com;foo.bar.baz'
```
mikkeloscar, line 265: We can define the typing here, so let's call this field `hostnames` and make it an list of strings so we don't need to encode the list in a single string.

pkg/core/autoscaler.go
```diff
@@ -284,6 +286,38 @@ func routegroupMetric(metrics zv1.AutoscalerMetrics, rgName, backendName string)
...
+	if metrics.Average == nil {
+		return nil, nil, fmt.Errorf("average value not specified for metric")
+	}
+
+	average := metrics.Average.DeepCopy()
+	generated := &autoscaling.MetricSpec{
+		Type: autoscaling.ExternalMetricSourceType,
+		External: &autoscaling.ExternalMetricSource{
+			Metric: autoscaling.MetricIdentifier{
+				Name: fmt.Sprintf("%s-rps", stacksetName),
```
AlexanderYastrebov, line 299: Will this work for multiple stacksets with the same name in different namespaces?

pkg/core/autoscaler.go
```diff
@@ -284,6 +286,38 @@ func routegroupMetric(metrics zv1.AutoscalerMetrics, rgName, backendName string)
...
+				},
+			},
+			Target: autoscaling.MetricTarget{
+				Type:         autoscaling.AverageValueMetricType,
+				AverageValue: &average,
+			},
+		},
+	}
+	annotations := map[string]string{
+		fmt.Sprintf("metric-config.%s-rps.requests-per-second/hostnames", stacksetName): metrics.Hostname,
```
AlexanderYastrebov, line 311: https://github.com/zalando-incubator/stackset-controller/pull/501/files#r1241866801

pkg/apis/zalando.org/v1/types.go
szuecs: @mikkeloscar I guess this would change the public interface which is in use by at least zalando applications.

docs/howtos.md
lucastt: > Also it looks like multiple hostnames should be separated by comma, not semicolon, shouldn't they?

Indeed. I'll fix it.

> Maybe we could make a bit clear how hostname (esp. when multiple are specified) and weight are related.

Sure. Actually I don't know if it makes much sense to have an interaction between weight and multiple hostnames. In my head at least weight would make more sense if you have multiple metrics. Because the way it works is that it multiplies the average by the percentage. When you have multiple hostnames the metric will just add all together and multiply for the percentage, so it would not really interact in any different way regardless of how many hostnames you have in the configuration.

Now rethinking this I start to realize that maybe this is not how weight was intended to be implemented. @mikkeloscar do you have any opinion regarding this?

pkg/core/autoscaler.go
lucastt: I believe yes, this part of the code creates a `MetricSpec` element that will be inserted into a `MetricSpec` array under an HPA object. As far as I understand the HPA object would be created in the namespace specified in the stackset object and there is no reference to any other k8s object, just hostnames that will be used in a prometheus query.

pkg/core/autoscaler.go
lucastt: Same answer as above, this should not be a problem because this is bound to an HPA resource that will exist only in the namespace the stackset is in. There is no reference to any other k8s object, just hostnames that will be used in a prometheus query.

pkg/core/autoscaler.go
mikkeloscar: Technically the API allows defining _multiple_ RPS based scaling metrics for a single StackSet, think about this case when generating a name, it needs to be unique within a single HPA resource.

pkg/core/autoscaler.go
```diff
@@ -284,6 +286,48 @@ func routegroupMetric(metrics zv1.AutoscalerMetrics, rgName, backendName string)
...
+				Type:         autoscaling.AverageValueMetricType,
+				AverageValue: &average,
+			},
+		},
+	}
+
+	hosts := ""
+	for _, h := range metrics.Hostnames {
+		hosts += h + ","
+	}
```
mikkeloscar, line 318: [`strings.Join`](https://pkg.go.dev/strings#example-Join)? :)

pkg/core/autoscaler.go
```diff
@@ -284,6 +286,48 @@ func routegroupMetric(metrics zv1.AutoscalerMetrics, rgName, backendName string)
...
+
+	annotations := map[string]string{
+		fmt.Sprintf("metric-config.%s-rps.requests-per-second/hostnames", stacksetName): hosts[:len(hosts)-1],
+	}
+
+	if metrics.Weight != "" {
+		annotations[fmt.Sprintf("metric-config.%s-rps.requests-per-second/weight", stacksetName)] = metrics.Weight
```
mikkeloscar, line 325: See also my other comment, the weight should be based on the traffic going to the particular stack. We do this for Ingress/RouteGroup metrics where you can see how we pass it.

mikkeloscar, line 325: As mentioned in another comment, the weight should be set based on the traffic weight of the Stack. We need to pass this information along when creating the HPA as we currently don't pass it iirc.

docs/howtos.md
lucastt: Hmm this is a good idea, I just don't know how to get the stack traffic information, I'll dig a little bit.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -266,6 +266,15 @@ type MetricsClusterScalingSchedule struct {
 	Name string `json:"name"`
 }
 
+// MetricRequestsPerSecond specifies basic information to scale based on
+// on external RPS metric.
+// +k8s:deepcopy-gen=true
+type MetricsRequestsPerSecond struct {
+	Name      string   `json:"name"`
+	Hostnames []string `json:"hostnames"`
+	Weight    string
```
lucastt, line 275: This has no json annotations, this is on purpose. I left this field here just to not break the tests, next commit I intend to remove this field and get weight from the Stack.

docs/howtos.md
lucastt: The change is done, now `weight` is not a parameter for the autoscaler metric, instead we get `weight` from the stack property `actualTrafficWeight`, this is passed to the HPA fabric function that passes the value along to the function responsible to generate the HPA metric resource. Please take a look and let me know what you think. I'm pondering if I should do the same with hosts, do you think it makes sense?

pkg/apis/zalando.org/v1/types.go
```diff
@@ -266,6 +266,14 @@ type MetricsClusterScalingSchedule struct {
 	Name string `json:"name"`
 }
 
+// MetricRequestsPerSecond specifies basic information to scale based on
+// on external RPS metric.
+// +k8s:deepcopy-gen=true
+type MetricsRequestsPerSecond struct {
+	Name      string   `json:"name"`
```
mikkeloscar, line 273: We don't expose a name in other metric types so I would also not do it here. It's just more for the users to write and we only need the hostnames defined to produce the rest for the HPA.

docs/howtos.md
mikkeloscar: For hosts this only works if they are using Ingress or RouteGroup where we already have the hostnames, but for those use cases using RouteGroup or Ingress is already the way to go so I would not see a need to change that unless we want to make all RPS use cases use the RequestsPerSecond type in the future in favor of Ingress and RouteGroup.
IMO we don't need this at this point. I would rather make it clear in the docs that this is intended for cases where you use externalIngress i.e. FabricGateway.

lucastt: Cool, it makes sense. I'll then update the docs and make the PR ready to merge.

pkg/apis/zalando.org/v1/types.go
mikkeloscar: Actually we _don't_ solve this problem for other metrics right now (I thought we did) so I would say it's ok to be out of scope for now.

My purpose would be that we simply pre/suffix a number to the name such that they become unique and stay small. We iterate through the metrics and can just carry and index number and use that. But let's consider it outside of scope for now and simply use a static name, like you already had in the original code.

lucastt: Yeah, I thought about it too, the only problem I have with this is that we can't be sure that metrics will always be in that index and I don't know the side effects of changing the name when someone changes the order or add/remove some metric, I didn't dive too much into the code and I don't know if we would not endup with more than one HPA for the same thing and cause problems because of that.

Anyway, I'll consider this out of scope as you suggested and remove this property.

docs/howtos.md
```diff
@@ -250,6 +251,30 @@ autoscaler:
     average: 30
 ```
 
+If the operator does not wish to reference a `Ingress` or `RouteGroup` resource for some reason like RouteGroups are 
+being generated by `FabricGateway`, and thus its hard to actually find which resource to reference using other metrics,
+a external metric like `RequestsPerSecond` is recommended, see the example:
```
mikkeloscar, line 256: I think it's a bit hard to understand. I would just say if you use `externalIngress` then this is the way to do RPS based scaling.

docs/howtos.md
```diff
@@ -250,6 +251,30 @@ autoscaler:
...
+  maxReplicas: 3
+  metrics:
+  - type: RequestsPerSecond
+    average: 30
+    requestsPerSecond:
+      hostnames: 
+        - 'example.com'
+        - 'foo.bar.baz'
+```
+When generating the HPA the autoscaler fabric will user the current traffic wait of the stack, for example: lets say the a traffic switch is happening from `stack-A` to `stack-B`, in the current state 30% of the traffic is being routed to `stack-B` backends, when calculating the average the metric will get the total traffic to `example.com` and `foo.bar.baz` and sum it all up, after this the final value is multiplied by the percentage weight in this case 30% resulting in something like:
```
mikkeloscar, line 270: The first sentence is a bit hard to understand. Instead of say "When generating the HPA", which is implementation detail, just state that the RPS of the hostname(s) is relative to the traffic that the stack is getting, which you explain in the next sentences.

pkg/core/autoscaler.go
```diff
@@ -284,6 +286,47 @@ func routegroupMetric(metrics zv1.AutoscalerMetrics, rgName, backendName string)
...
+	if len(metrics.RequestsPerSecond.Hostnames) == 0 {
+		return nil, nil, fmt.Errorf("RequestsPerSecond.hostnames value not specified for metric")
+	}
+
+	average := metrics.Average.DeepCopy()
+	generated := &autoscaling.MetricSpec{
+		Type: autoscaling.ExternalMetricSourceType,
+		External: &autoscaling.ExternalMetricSource{
+			Metric: autoscaling.MetricIdentifier{
+				Name: stackname,
```
mikkeloscar, line 307: Let's maybe at least add something like `rps` to the name so you understand it's an rps metric from looking at the name.

lucastt: True, I'm sorry

docs/howtos.md
MustafaSaber, line 270: `will user the current traffic` to `will use the current traffic`
`lets say the a traffic` to `let's say the traffic`
I would suggest using number other than 30% in the traffic weight since it may cause confusion with the `average: 30`

lucastt:
👍

pkg/core/autoscaler.go
```diff
@@ -299,12 +299,14 @@ func externalRPSMetric(metrics zv1.AutoscalerMetrics, stackname string, weight f
...
 
+	name := stackname + "-rps"
+
 	average := metrics.Average.DeepCopy()
 	generated := &autoscaling.MetricSpec{
 		Type: autoscaling.ExternalMetricSourceType,
 		External: &autoscaling.ExternalMetricSource{
 			Metric: autoscaling.MetricIdentifier{
-				Name: stackname,
+				Name: name,
```
mikkeloscar, line 309: You need to also use this variable on line 321 and 322

lucastt: my bad

lucastt: I honestly don't know how the tests did not get it 🤔

mikkeloscar: The test checks for the old name as well 😅

mikkeloscar:
:+1:

lucastt:
👍

PR #502, opened 2023-06-28
fix cpu metric in readme

dneuhaeuser-zalando:
This needs to be uppercase, using `cpu` lowercase results in:

```
spec.stackTemplate.spec.autoscaler.metrics.type: Unsupported value: "cpu": supported values: "CPU", "Memory", "AmazonSQS", "PodJSON", "Ingress", "RouteGroup", "ZMON", "ScalingSchedule", "ClusterScalingSchedule"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ spec:
         minReplicas: 3
         maxReplicas: 10
         metrics:
-        - type: cpu
+        - type: CPU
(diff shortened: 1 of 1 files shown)


dneuhaeuser-zalando:
👍

mikkeloscar:
:+1:

PR #511, opened 2023-07-13
Add RequestsPerSecond type to validation

mikkeloscar:
Another follow up to #501 adding the `RequestsPerSecond` type in the CRD validation. This is required to avoid the error:

```
The StackSet "x" is invalid: spec.stackTemplate.spec.autoscaler.metrics.type: Unsupported value: "RequestsPerSecond": supported values: "CPU", "Memory", "AmazonSQS", "PodJSON", "Ingress", "RouteGroup", "ZMON", "ScalingSchedule", "ClusterScalingSchedule"
```

### DIFF
--- a/docs/stack_crd.yaml
+++ b/docs/stack_crd.yaml
@@ -294,6 +294,7 @@ spec:
                           - ZMON
                           - ScalingSchedule
                           - ClusterScalingSchedule
+                          - RequestsPerSecond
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #518, opened 2023-08-01
Include ingress and/or routegroup definitions in Stack template.

gargravarr:
This Pull Request (PR) updates the following:

* adds to the Stack CRDs the Ingress, RouteGroup and ExternalIngress definitions. In this way we can keep these definitions in each Stack scope.
* Removes the Ingress/RouteGroup overrides per stack, so far not used. We will have to find another solution for the problem described here https://github.com/zalando-incubator/stackset-controller/pull/329
* When the reconciler creates a new stack: when existing, overrides the ingress, routegroup  and external ingress spec of the parent StackSet to the new Stack spec. This is backward compatible to the current stacks, because it's an _override_: stacks that don't have ingress/routegroup spec will just use the parent stackset spec, in case of deletion of ingress/routegroup.


pkg/core/types.go
```diff
@@ -355,8 +356,38 @@ func (ssc *StackSetContainer) TrafficChanges() []TrafficChange {
...
+	overridePort := false
+	if sc.Stack.Spec.Ingress != nil {
+		overridePort = true
+		sc.ingressSpec = sc.Stack.Spec.Ingress
+	}
+
+	if sc.Stack.Spec.RouteGroup != nil {
+		overridePort = true
+		sc.routeGroupSpec = sc.Stack.Spec.RouteGroup
+	}
```
mikkeloscar, line 371: Don't you need to also check for `ExternalIngress` here? Otherwise if only `ExternalIngress` is defined on the stack then `overridePort` will be false.

gargravarr: Right, I missed that. For `ExternalIngress` the only override is the port.

gargravarr: Fixed

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #519, opened 2023-08-02
Add stack weight verification for HPA reconciling

lucastt:
This PR introduces a new check to make weight changes to External RPS HPA automatically during traffic switch. Now when traffic switching using stackset weights the weight change trigger the update of the HPA that will take in consideration the new weight of the stack. Before only stack generation was considered, which would not be updated in Status and annotation changes.

### DIFF
--- a/controller/stack_resources.go
+++ b/controller/stack_resources.go
@@ -27,24 +27,24 @@ func pint32Equal(p1, p2 *int32) bool {
 // e.g. External RPS metric, this verification provides a way to verify
 // all relevant annotations are actually up to date.
 func areHPAAnnotationsUpToDate(updated, existing *v2.HorizontalPodAutoscaler) bool {
-    if len(updated.Annotations) != len(existing.Annotations) {
-        return false
-    }
+	if len(updated.Annotations) != len(existing.Annotations) {
+		return false
+	}
 
-    for k, v := range updated.Annotations {
-        if k == "stackset-controller.zalando.org/stack-generation" {
-            continue
-        }
+	for k, v := range updated.Annotations {
+		if k == "stackset-controller.zalando.org/stack-generation" {
+			continue
+		}
 
-        existingValue, ok := existing.Annotations[k]
-        if ok && existingValue == v {
-            continue
-        }
+		existingValue, ok := existing.Annotations[k]
+		if ok && existingValue == v {
+			continue
+		}
 
-        return false
-    }
+		return false
+	}
 
-    return true
+	return true
 }
 
 // syncObjectMeta copies metadata elements such as labels or annotations from source to target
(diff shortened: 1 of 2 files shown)


controller/stack_resources.go
```diff
@@ -23,6 +25,32 @@ func pint32Equal(p1, p2 *int32) bool {
 	return false
 }
 
+func isHPATrafficWeightUpdated(stack *zv1.Stack, hpa *v2.HorizontalPodAutoscaler) (bool, error) {
```
mikkeloscar, line 28: I think this is the right idea, but it becomes very specific to this particular case. The `ReconcileStackHPA()` method is intentionally quite generic it just takes a function for generating an HPA and checks if it needs an update. With this in mind I think we can also make this check more generic. What we really want is just to check if annotations changed between the current version and the new generated HPA.

My proposal would be to simply compare the annotations (excluding the special `stackset-controller.zalando.org/stack-generation` annotation) and if there is a diff we know we need to update the HPA. That way we don't need a very custom check if weight changed.

controller/stack_resources.go
```diff
@@ -23,6 +23,26 @@ func pint32Equal(p1, p2 *int32) bool {
...
+        existingValue, ok := existing.Annotations[k]
+        if ok && existingValue == v {
+            continue
+        }
+
+        return false
+    }
+
+    return true
+}
```
mikkeloscar, line 44: I think this has some uncovered edge cases like: One annotation was removed from the updated hpa.

lucastt: Good catch. I fixed it by also comparing the length this way we only return true if annotation len is equal and every annotation present in on of the maps is also present in the other with the same value.

mikkeloscar: It almost works :sweat_smile: The generated HPA won't have the `stackset-controller.zalando.org/stack-generation` annotation so the equal length check will always be false, which makes it work, but means the HPA would _always_ be updated. It needs to be slightly more sophisticated to exclude `stackset-controller.zalando.org/stack-generation` and then compare length and check for key/value.

lucastt: > The generated HPA won't have the stackset-controller.zalando.org/stack-generation annotation so the equal length check will always be false

Oh, true 😢

lucastt: Wait, the generated HPA also has stack generation, I thought it was added afterwards but:

https://github.com/zalando-incubator/stackset-controller/blob/external-rps-hpa-reconcile/pkg/core/stack_resources.go#L83

and `StackContainer.resourceMeta()` is called by `StackContainer.GenerateHPA()` that is the function that generates the HPA:

https://github.com/zalando-incubator/stackset-controller/blob/external-rps-hpa-reconcile/pkg/core/stack_resources.go#L224

So both should have `stackset-controller.zalando.org/stack-generation` at this point. Did I get something wrong?

If this is not the case the tests for this component are wrong because some of them assume `stackset-controller.zalando.org/stack-generation` is set in the generated HPA. 🤔

mikkeloscar: You're right, sorry that I let you in the wrong direction :(

lucastt: No worries :)

lucastt:
👍

PR #520, opened 2023-08-22
Reference ConfigMap for per stack versioning

katyanna:


cmd/e2e/basic_test.go
```diff
@@ -156,12 +156,11 @@ func (f *TestStacksetSpecFactory) Create(stackVersion string) zv1.StackSetSpec {
 		}
 
 		if f.hpaBehavior {
-			result.StackTemplate.Spec.HorizontalPodAutoscaler.Behavior =
-				&autoscalingv2.HorizontalPodAutoscalerBehavior{
-					ScaleDown: &autoscalingv2.HPAScalingRules{
-						StabilizationWindowSeconds: &f.hpaStabilizationWindowSeconds,
-					},
-				}
+			result.StackTemplate.Spec.HorizontalPodAutoscaler.Behavior = &autoscalingv2.HorizontalPodAutoscalerBehavior{
```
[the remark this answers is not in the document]
szuecs: or maybe drop this unrelated change :)

linki:
I'm breaking up this PR in smaller manageable pieces:
1) CRD spec: https://github.com/zalando-incubator/stackset-controller/pull/528

linki:
Attempt to break this up into smaller PRs.
- [ ] 1. CRD spec: https://github.com/zalando-incubator/stackset-controller/pull/528
- [ ] 2. Basic Implementation: https://github.com/zalando-incubator/stackset-controller/pull/529
- [ ] 3. Test cases

katyanna:
@linki Given the atomicity of the feature and since the derived pull requests don't add actual changes in relation to this one, I don't really see a need for multiple PRs. The commits separate implementation, CRD definition, and tests to facilitate review as we aligned yesterday.

cmd/e2e/basic_test.go
```diff
@@ -345,6 +346,13 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 		require.EqualValues(t, stackRGBackends, stackRG.Spec.Backends)
 		require.EqualValues(t, stackRGHosts, stackRG.Spec.Hosts)
 	}
+
+	// Verify the ConfigMaps
+	if stacksetSpec.StackTemplate.Spec.ConfigResources != nil {
+		configMap, err := waitForConfigMap(t, stack.Name)
+		require.NoError(t, err)
+		require.EqualValues(t, stackResourceLabels, configMap.Labels)
```
linki, line 354: We should probably check for more things here, e.g. the content of the ConfigMap.

cmd/e2e/test_utils.go
```diff
@@ -440,6 +440,14 @@ func waitForRouteGroup(t *testing.T, name string) (*rgv1.RouteGroup, error) {
 	return routegroupInterface().Get(context.Background(), name, metav1.GetOptions{})
 }
 
+func waitForConfigMap(t *testing.T, name string) (*corev1.ConfigMap, error) {
+	err := resourceCreated(t, "config-map", name, configMapInterface()).await()
```
linki, line 444: This should be `"configmap"`

controller/stack_resources.go
```diff
@@ -318,3 +318,68 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+	existing *apiv1.ConfigMap,
+	generateUpdated func(*apiv1.ConfigMap) (*apiv1.ConfigMap, error),
+) error {
+	if stack.Spec.ConfigResources == nil || existing != nil {
+		return nil
+	}
+
+	// Generate versioned ConfigMap object based on referenced template
+	templateName := stack.Spec.ConfigResources.ConfigMapRef.Name
+	template, err := c.client.CoreV1().ConfigMaps(stack.Namespace).Get(ctx, templateName, metav1.GetOptions{})
```
linki, line 334: We could think about moving this part into `generateUpdated()`.

Similar to `ReconcileStackRouteGroup` and others, `generateUpdated()` is called on the first line of the function and returns the _final_ resource, which is then merely deleted, created or updated (if changed) in this function. There's no "business logic" in the other functions, so we could think about doing that here too.

controller/stack_resources.go
```diff
@@ -318,3 +318,68 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+	}
+	c.recorder.Eventf(
+		stack,
+		apiv1.EventTypeNormal,
+		"CreatedConfigMap",
+		"Created ConfigMap %s",
+		configMap.Name)
+
+	// Update ConfigMap to be mounted on Stack's pods
+	for _, scm := range stack.Spec.PodTemplate.Spec.Volumes {
```
linki, line 356: We also need to handle `envFrom`.

controller/stack_resources.go
```diff
@@ -318,3 +318,68 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+		if scm.ConfigMap.Name == templateName {
+			scm.ConfigMap.Name = configMap.Name
+		}
+	}
+	_, err = c.client.ZalandoV1().Stacks(stack.Namespace).Update(ctx, stack, metav1.UpdateOptions{})
```
linki, line 361: We need to double-check whether we it's safe to update `Stacks` from this function. The other `Reconcile*` functions solely manage their own resource (`ReconcileStackDeployment` for the `Deployment` object, `ReconcileStackHPA` for the `HPA` object etc.). This function works similar for `ConfigMaps` but this line also patches the `Stack` which is something we didn't do before in that sort of function.

controller/stack_resources_test.go
```diff
@@ -956,3 +956,60 @@ func TestReconcileStackRouteGroup(t *testing.T) {
...
+                Data: map[string]string {
+                    "testK": "testV",
+                },
+                Immutable: nil,
+			},
+			expected: &v1.ConfigMap {
+				ObjectMeta: baseTestStackOwned,
+                Data: map[string]string {
+                    "testK": "testV",
+                },
```
linki, line 982: Formatting errors?

controller/stackset.go
```diff
@@ -428,6 +433,27 @@ Items:
...
+		if uid, ok := getOwnerUID(configMap.ObjectMeta); ok {
+			for _, stackset := range stacksets {
+				if s, ok := stackset.StackContainers[uid]; ok {
+					s.Resources.ConfigMap = &configMap
+					continue Items
+				}
+			}
+		}
+	}
+	return nil
```
linki, line 454: Can you explain the use of `continue Items`. I see that `collectHPAs` also uses it but `collectDeployments` using `break` looks much more reasonable.

docs/stack_crd.yaml
```diff
@@ -347,6 +347,23 @@ spec:
...
+                    properties:
+                      name:
+                        description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                          TODO: Add other useful fields. apiVersion, kind, uid?'
+                        type: string
+                      optional:
+                        description: Specify whether the ConfigMap must be defined
+                        type: boolean
+                    type: object
+                type: object
```
linki, line 366: This spec should aligned with what we have in the design document: https://docs.google.com/document/d/1qboqAKuqfUOx7VZr_V9SFq5n6r5yBVXibRTZFapvCPY/edit?usp=sharing

pkg/apis/zalando.org/v1/types.go
```diff
@@ -428,6 +428,17 @@ type StackSpec struct {
 
 	// Strategy describe the rollout strategy for the underlying deployment
 	Strategy *appsv1.DeploymentStrategy `json:"strategy,omitempty"`
+
+	// ConfigResources describes the ConfigMaps that will be created.
+	// Later Secrets and PlatformCredentialSets will also be defined on ConfigResources
+	ConfigResources *[]ConfigResourcesSpec `json:"configResources,omitempty"`
```
katyanna, line 434: In the document, we used `TemplateResources` instead. Since we plan to work with inline definition also that other resources (HPAs, Ingresses, etc) will not be included in this section, I believe `ConfigResources` may be a clearer option.

Another alternative would be `ConfigurationResources`. Please let me know your thoughts.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -428,6 +428,17 @@ type StackSpec struct {
...
+}
+
+// ConfigResourcesSpec makes it possible to defined the config resources to be created
+// +k8s:deepcopy-gen=true
+type ConfigResourcesSpec struct {
+	// ConfigMap to be versioned for Stack
+	ConfigMapRef v1.ConfigMapEnvSource `json:"configMapRef,omitempty"`
```
katyanna, line 441: In the design document, we had a **Name** field here.
```
 stackTemplate:
    spec:
      version: v1
      configResources:
      - name: my-config
        configMapRef:
          name: app-config
```

However this information was not used anywhere and I believe may cause confusion as the user might expect that to be the final name of the versioned resource, when that's actually set automatically as `<stack-name>-<template-name>`.

We could still add it and use in the final name as `<stack-name>-<given-name>`. I just don't see the value at the moment, please let me know what you think.

controller/stack_resources.go
katyanna: In my understanding, the work to be done in the reconcile functions is to reconcile Stack<>Resource, so far I haven't found any indication that updating the Stack isn't advised here.

controller/stack_resources.go
katyanna: In the `Generate<Resource>` methods we don't have straight access to the controller, which is needed to get the `ConfigMap` template object we'll work on. So for that to be passed on to the generation of the versioned ConfigMap, we have some work to do beforehand in the Reconcile.

That is not the case in the previous implemented resources because they have all the data they need in the `Stack`.

pkg/apis/zalando.org/v1/types.go
zaklawrencea: +1 for `ConfigurationResources`. As a bonus it also aligns with the Kubernetes docs [Concepts -> Configuration](https://kubernetes.io/docs/concepts/configuration/).

linki: Sounds good to me 👍

pkg/apis/zalando.org/v1/types.go
linki: It actually is inspired by how Kubernetes does it, e.g. for volumes Kubernetes uses this syntax:
```yaml
  volumes:
  - name: config
    configMap:
      name: game-demo
```
I assume the intention was to decouple the name within the Pod (`config`) from what is actually backing it (the ConfigMap called `game-demo`).

I'm fine with dropping the name field here (or make it optional and infer a default from the ConfigMap name) but as far as I can see we will need it for the inline solution:
```yaml
 stackTemplate:
    spec:
      version: v1
      configResources:
      - name: my-config
        configMap:
          key1: value1
```
Without the name there's no way to refer to this config in the PodSpec.

controller/stackset.go
linki: This is just to `break` the outer loop from the inside, similar to described here: https://relistan.com/continue-statement-with-labels-in-go (A better way).

Since the label is _before_ the loop it looks like it would re-execute the entire loop for all items again, but I guess it just continues with the next item of the range.

controller/stack_resources.go
linki: Let's see how it works for the early access. My fear is that the additional stack updates (it's one per configmap) can somehow lead to more race conditions and/or errors. The only alternative would be to move that part to `ReconcileStackDeployment` and then only rewrite the config map references on creation of the Deployment object.

But the current logic is correct, so that's not blocking us from trying it out.

cmd/e2e/basic_test.go
```diff
@@ -345,6 +346,13 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 		require.EqualValues(t, stackRGBackends, stackRG.Spec.Backends)
 		require.EqualValues(t, stackRGHosts, stackRG.Spec.Hosts)
 	}
+
+	// Verify the ConfigMaps
+	if stacksetSpec.StackTemplate.Spec.ConfigurationResources != nil {
```
linki, line 351: We need to revisit the remaining places where we check for `ConfigurationResources != nil`. Since it's a list now that is most likely not correct. For example here, it only checks for a single configmap (`waitForConfigMap(t, stack.Name)`) when the stack has at least one config resources (`ConfigurationResources != nil`). That doesn't seem correct.

controller/stack_resources.go
```diff
@@ -318,3 +323,127 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+				}
+			}
+		}
+
+		// Change ConfigMap reference on Stack's Volumes
+		for _, volume := range stack.Spec.PodTemplate.Spec.Volumes {
+			if volume.ConfigMap.Name == templateName {
+				volume.ConfigMap.Name = versionedName
+			}
+		}
```
linki, line 363: If we want we could also already add the _third_ way of using configmaps, via `env[*].valueFrom`: https://github.bus.zalan.do/teapot/admission-controller/pull/204/files#diff-e926f96370935b91241e7a36cb31ecace7eea344be6fda8ccb70e330b9103fedR158-R164

controller/stack_resources.go
```diff
@@ -318,3 +323,127 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+	// Get configMaps names
+	configMaps := make(map[string]string)
+	for _, configMap := range *stack.Spec.ConfigurationResources {
+		templateName := configMap.ConfigMapRef.Name
+		configMaps[templateName] = generateConfigMapName(stack, templateName)
+	}
+
+	if len(existing) >= len(configMaps) {
+		for templateName := range configMaps {
+			err := c.deleteConfigMapTemplate(ctx, stack, templateName)
```
linki, line 401: When does this code run? I see that we also have `deleteConfigMapTemplate` further below after each configitem was processed. Why do we delete multiple times?

pkg/core/stack_resources.go
```diff
@@ -421,6 +421,25 @@ func (sc *StackContainer) GenerateRouteGroup() (*rgv1.RouteGroup, error) {
 	return result, nil
 }
 
+func (sc *StackContainer) GenerateConfigMap(base *v1.ConfigMap, name string) (*v1.ConfigMap, error) {
+	stackSpec := sc.Stack.Spec
+	metaObj := sc.resourceMeta()
+	metaObj.Name = name
+
+	if stackSpec.ConfigurationResources != nil {
```
linki, line 429: I think we can remove that `if`. If we are inside this function then `ConfigurationResources != nil` is always true, isn't it?

controller/stack_resources.go
katyanna: This tends specially to the runs after the first. If the Stack already has the same amount of versioned ConfigMaps as defined in the manifest, then this deletes the ConfigMap templates and exits the method. If not, the method moves on to the creation of the missing resources.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -428,6 +428,17 @@ type StackSpec struct {
...
+	// ConfigurationResources describes the ConfigMaps that will be created.
+	// Later Secrets and PlatformCredentialSets will also be defined on ConfigurationResources
+	ConfigurationResources *[]ConfigurationResourcesSpec `json:"configResources,omitempty"`
+}
+
+// ConfigurationResourcesSpec makes it possible to defined the config resources to be created
+// +k8s:deepcopy-gen=true
+type ConfigurationResourcesSpec struct {
+	// ConfigMap to be versioned for Stack
+	ConfigMapRef v1.ConfigMapEnvSource `json:"configMapRef,omitempty"`
```
linki, line 441: We should use `v1.LocalObjectReference` instead of `v1.ConfigMapEnvSource`, like here: https://github.com/zalando-incubator/stackset-controller/pull/528/commits/07f39ba81d4e2b1b22f7f9b218047ac8006647ca#diff-f37f26db53b0d98af402010451153f66da4061eb531e4ca1665086efb26ef0fdR454

`v1.ConfigMapEnvSource` is a "subclass" of `v1.LocalObjectReference` that adds the `optional` field, which we don't want. See the generated StackSet CRD spec:

image

pkg/apis/zalando.org/v1/types.go
```diff
@@ -428,6 +428,17 @@ type StackSpec struct {
 
 	// Strategy describe the rollout strategy for the underlying deployment
 	Strategy *appsv1.DeploymentStrategy `json:"strategy,omitempty"`
+
+	// ConfigurationResources describes the ConfigMaps that will be created.
+	// Later Secrets and PlatformCredentialSets will also be defined on ConfigurationResources
+	ConfigurationResources *[]ConfigurationResourcesSpec `json:"configResources,omitempty"`
```
linki, line 434: Usually it's not necessary to use a pointer to a slice, even when it's optional. The slice is already a pointer. This should also work fine:

```go
ConfigurationResources []ConfigurationResourcesSpec
```

controller/stackset.go
```diff
@@ -269,6 +282,11 @@ func (c *StackSetController) collectResources(ctx context.Context) (map[types.UI
 		return nil, err
 	}
 
+	err = c.collectConfigMaps(ctx, stacksets)
```
linki, line 285: Let's also put that behind the feature flag. This way we reduce the amount of code changes while the feature is still turned of (production clusters until after CW).

controller/stack_resources.go
```diff
@@ -318,3 +323,139 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+func (c *StackSetController) updateStackConfigMap(
+	ctx context.Context,
+	stack *zv1.Stack,
+	configMapNames map[string]string,
+) error {
+	for templateName, versionedName := range configMapNames {
+		// Change ConfigMap reference on Stack's EnvFrom
+		for _, container := range stack.Spec.PodTemplate.Spec.Containers {
+			for _, envFrom := range container.EnvFrom {
+				if envFrom.ConfigMapRef.Name == templateName {
```
linki, line 352: This will `panic` when `envFrom.ConfigMapRef` is `nil`, for example when doing this:
```yaml
envFrom:
  secretRef: foo
```

We need to check for `envFrom.ConfigMapRef != nil && envFrom.ConfigMapRef.Name == templateName`, like here: https://github.bus.zalan.do/teapot/admission-controller/pull/204/files#diff-e926f96370935b91241e7a36cb31ecace7eea344be6fda8ccb70e330b9103fedR147

controller/stack_resources.go
```diff
@@ -318,3 +323,139 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+					envFrom.ConfigMapRef.Name = versionedName
+				}
+			}
+		}
+
+		// Change ConfigMap reference on Stack's Volumes
+		for _, volume := range stack.Spec.PodTemplate.Spec.Volumes {
+			if volume.ConfigMap.Name == templateName {
```
linki, line 360: Similar to the case for `envFrom`, this panics when using any other volume type. We need to check that `ConfigMap` is non-nil before querying the `Name` field (see also: https://github.bus.zalan.do/teapot/admission-controller/pull/204/files#diff-e926f96370935b91241e7a36cb31ecace7eea344be6fda8ccb70e330b9103fedR135)

pkg/core/stack_resources_test.go
```diff
@@ -1049,3 +1049,59 @@ func TestGenerateStackStatus(t *testing.T) {
...
+			name: "ConfigMap name is updated",
+			template: &v1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "configmap",
+					Namespace: testStackMeta.Namespace,
+				},
+				Data: map[string]string{
+					"testK": "testV",
+				},
+				Immutable: nil,
```
linki, line 1077: As discussed, let's remove the immutable "feature". The users prefer "normal" ConfigMaps for emergency handling.

linki:
I added a couple of remarks that I think we should fix for the final version but we can do that in a separate PR.

To get this merged, let's do [this comment](https://github.com/zalando-incubator/stackset-controller/pull/520#pullrequestreview-1692088921) (putting `collectConfigMaps` behind feature flag) and [this problem](https://github.com/zalando-incubator/stackset-controller/pull/520#pullrequestreview-1692149706) (avoiding panics during configmap name rewrite).

pkg/apis/zalando.org/v1/types.go
```diff
@@ -438,7 +438,7 @@ type StackSpec struct {
 // +k8s:deepcopy-gen=true
 type ConfigurationResourcesSpec struct {
 	// ConfigMap to be versioned for Stack
-	ConfigMapRef v1.LocalObjectReference `json:"configMapRef,omitempty"`
+	ConfigMapRef *v1.LocalObjectReference `json:"configMapRef,omitempty"`
```
linki, line 441: Let's keep it like before, without the `*` if possible.

pkg/core/stack_resources_test.go
```diff
@@ -1092,12 +1090,12 @@ func TestGenerateConfigMap(t *testing.T) {
 				Data: map[string]string{
 					"testK": "testV",
 				},
-				Immutable: &immutable,
+				Immutable: nil,
```
linki, line 1093: Let's just remove this field entirely. I think the default is `nil` anyways.

controller/stack_resources.go
```diff
@@ -324,144 +320,53 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
 func (c *StackSetController) ReconcileStackConfigMap(
 	ctx context.Context,
 	stack *zv1.Stack,
 	existing []*apiv1.ConfigMap,
-	generateUpdated func(*apiv1.ConfigMap, string) (*apiv1.ConfigMap, error),
+	generateUpdated func(*apiv1.ConfigMap) (*apiv1.ConfigMap, error),
 ) error {
 	if stack.Spec.ConfigurationResources == nil {
 		return nil
 	}
```
linki, line 337: If think we can remove this now. The `len()` and the `for` loop below will just work fine. `len` returns `0` and the `for` loop will be skipped.

pkg/core/stack_resources.go
```diff
@@ -421,19 +421,13 @@ func (sc *StackContainer) GenerateRouteGroup() (*rgv1.RouteGroup, error) {
...
 }
 
-func (sc *StackContainer) GenerateConfigMap(base *v1.ConfigMap, name string) (*v1.ConfigMap, error) {
+func (sc *StackContainer) GenerateConfigMap(configMap *v1.ConfigMap) (*v1.ConfigMap, error) {
 	metaObj := sc.resourceMeta()
-	metaObj.Name = name
-	metaObj.Annotations = mergeLabels(metaObj.Annotations, base.Annotations)
+	configMap.OwnerReferences = metaObj.OwnerReferences
+	configMap.Labels = metaObj.Labels
+	configMap.Annotations = mergeLabels(metaObj.Annotations, configMap.Annotations)
```
linki, line 428: I'm not sure if we want to modify the Labels and Annotations of the referenced ConfigMap in the revised approach. It would also _overwrite_ any user supplied labels I think (`configMap.Labels = metaObj.Labels`). Let's just leave it out for now and only do the ownerReference.

linki: The `stack-version` and `stackset` labels make sense.

controller/stack_resources.go
```diff
@@ -318,3 +319,55 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+	if stack.Spec.ConfigurationResources == nil {
+		return nil
+	}
+
+	if len(existing) >= len(stack.Spec.ConfigurationResources) {
+		return nil
+	}
+
+	for _, rsc := range stack.Spec.ConfigurationResources {
+		configMap, err := c.client.CoreV1().ConfigMaps(stack.Namespace).Get(ctx, rsc.ConfigMapRef.Name, metav1.GetOptions{})
```
linki, line 344: If `rsc.ConfigMapRef.Name` doesn't follow the pattern `"{{{APPLICATION}}}-{{{VERSION}}}-configmap-name"` then we should return an error so that the Stack isn't created.

controller/stack_resources.go
katyanna: Not really, `len` will fail if `stack.Spec.ConfigurationResources == nil`.

pkg/core/stack_resources.go
katyanna: I'll do the `mergeLabels` to ensure we don't lose the labels from the ConfigMap, as it's done with the Annotations.

linki:
The ConfigMap name pattern check's error message doesn't yet show up in Sunrise.

image

controller/stack_resources.go
```diff
@@ -318,3 +344,71 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+		configMap, err := c.client.CoreV1().ConfigMaps(stack.Namespace).
+			Get(ctx, rscName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if configMap.OwnerReferences != nil {
+			for _, owner := range configMap.OwnerReferences {
+				if owner.UID != stack.UID {
+					return fmt.Errorf(`ConfigMap already owned by other resource.
```
linki, line 390: This adds a new line to the logline itself which looks very weird when lookign at the logs and the Kubernetes event. We need to find a way without the newline.

cmd/e2e/basic_test.go
```diff
@@ -488,12 +505,25 @@ func verifyStacksetRouteGroup(t *testing.T, stacksetName string, stacksetSpec zv
...
+	updateStrategy bool,
+	subResourceAnnotations map[string]string,
+) {
 	t.Parallel()
 
 	stacksetName := fmt.Sprintf("stackset-create-%s", testName)
 	stackVersion := "v1"
 	stacksetSpecFactory := NewTestStacksetSpecFactory(stacksetName)
+	if configmap {
+		stacksetSpecFactory.ConfigMap()
```
linki, line 525: Is this even doing anything? From looking at the code it just sets `f.configMap = true` but the field is never actually used anywhere.

cmd/e2e/basic_test.go
```diff
@@ -345,6 +353,15 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 		require.EqualValues(t, stackRGBackends, stackRG.Spec.Backends)
 		require.EqualValues(t, stackRGHosts, stackRG.Spec.Hosts)
 	}
+
+	// Verify ConfigMaps
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
```
linki, line 358: I don't think this list is ever filled with anything, making the checks inside the loop meaningless.

I've added a commit that flips the assertions. Some tests should fail as a consequence.

katyanna:
The immutability test was intentional. Its existence documents and ensures that immutability is not changed.

controller/stack_resources.go
```diff
@@ -318,3 +343,72 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
...
+
+		// ensure that ConfigurationResources are prefixed by Stack name.
+		if err := validateConfigurationResourceNames(stack); err != nil {
+			return err
+		}
+
+		configMap, err := c.client.CoreV1().ConfigMaps(stack.Namespace).
+			Get(ctx, rscName, metav1.GetOptions{})
+		if err != nil {
+			return err
```
linki, line 384: Let's think about this in the following PR: If the referenced ConfigMap **isn't found** we might want to simply skip it. This is especially interesting when you specify many ConigurationResources where some are missing.

In the new approach there's no need to copy from a template anymore, so we can just proceed and create the Stack. Kubernetes will then report errors when the referenced ConfigMaps don't exist.

If we look at what this function ultimately does: It just iterates over the list of referenced ConfigMaps and adds the ownerReference. If just one one of those cannot be found it will just stop completely, not processing the others. That doesn't seem necessary.

controller/stackset.go
```diff
@@ -604,6 +643,11 @@ func (c *StackSetController) CreateCurrentStack(ctx context.Context, ssc *core.S
 		return nil
 	}
 
+	// ensure that ConfigurationResources are prefixed by Stack name.
+	if err := validateConfigurationResourceNames(newStack.Stack); err != nil {
```
linki, line 647: Just a heads up: We might only want to validate when the feature is actually enabled 🤦  (`--enable-configmap-support`)

linki: I put it behind the feature flag.

linki:
👍

katyanna:
👍

PR #521, opened 2023-08-23
Drop HorizontalPodAutoscaler field

mikkeloscar:
This drops support for the legacy `horizontalPodAutoscaler` field and only allows the `autoscaler` field going forward.

The motivation for this change is that the `horizontalPodAutoscaler` field is based on autoscaler/v2beta1 API which is no longer allowed in Kubernetes v1.26 and it leads to confusion to support two different types of ways to define an autoscaler configuration.

This field is no longer in use by teams in Zalando


gargravarr:
:+1:

PR #537, opened 2023-10-19
Remove HPA when stack is scaled down

miladbarazandeh:
Teapot issue 3482

### DIFF
--- a/pkg/core/stack_resources.go
+++ b/pkg/core/stack_resources.go
@@ -223,6 +223,10 @@ func (sc *StackContainer) GenerateHPA() (*autoscaling.HorizontalPodAutoscaler, e
 		return nil, nil
 	}
 
+	if sc.ScaledDown() {
+		return nil, nil
+	}
+
 	result := &autoscaling.HorizontalPodAutoscaler{
 		ObjectMeta: sc.resourceMeta(),
 		TypeMeta: metav1.TypeMeta{
--- a/pkg/core/stack_resources_test.go
+++ b/pkg/core/stack_resources_test.go
@@ -895,6 +895,7 @@ func TestGenerateHPA(t *testing.T) {
 		autoscaler          *zv1.Autoscaler
 		expectedMinReplicas *int32
 		expectedMaxReplicas int32
+		noTrafficSince      time.Time
 		expectedMetrics     []autoscaling.MetricSpec
 		expectedBehavior    *autoscaling.HorizontalPodAutoscalerBehavior
 	}{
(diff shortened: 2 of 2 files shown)


controller/stack_resources.go
```diff
@@ -116,6 +117,20 @@ func (c *StackSetController) ReconcileStackHPA(ctx context.Context, stack *zv1.S
 		return nil
 	}
 
+	// Remove HPA if stackset is scaled down
+	if scaledDown {
+		err := c.client.AutoscalingV2().HorizontalPodAutoscalers(existing.Namespace).Delete(ctx, existing.Name, metav1.DeleteOptions{})
```
linki, line 122: `existing` must not be `nil` here, so I would move this below the next check.

controller/stack_resources.go
```diff
@@ -93,7 +94,7 @@ func (c *StackSetController) ReconcileStackDeployment(ctx context.Context, stack
 	return nil
 }
 
-func (c *StackSetController) ReconcileStackHPA(ctx context.Context, stack *zv1.Stack, existing *v2.HorizontalPodAutoscaler, generateUpdated func() (*v2.HorizontalPodAutoscaler, error)) error {
+func (c *StackSetController) ReconcileStackHPA(ctx context.Context, stack *zv1.Stack, existing *v2.HorizontalPodAutoscaler, generateUpdated func() (*v2.HorizontalPodAutoscaler, error), scaledDown bool) error {
 	hpa, err := generateUpdated()
```
linki, line 98: You can also investigate whether it makes sense to make the check for scaled down with this function. From what I see it also lives on the StackContainer object which has that information. You could then return `nil` for the `hpa` here and the rest of the function should take care of removing the `hpa` if it existed.

controller/stack_resources.go
```diff
@@ -116,6 +117,20 @@ func (c *StackSetController) ReconcileStackHPA(ctx context.Context, stack *zv1.S
...
+		if err != nil {
+			return err
+		}
+		c.recorder.Eventf(
+			stack,
+			apiv1.EventTypeNormal,
+			"DeletedHPA",
+			"Deleted HPA %s",
+			existing.Namespace)
+	}
```
linki, line 132: This code is exactly the same as the delete code above. Maybe we can combine them.

linki, line 132: We probably need to `return` after the deletion of the HPA. Otherwise the remainder of the function might run into a panic.

controller/stackset.go
```diff
@@ -972,7 +972,7 @@ func (c *StackSetController) ReconcileStackResources(ctx context.Context, ssc *c
 	if err != nil {
 		return c.errorEventf(sc.Stack, "FailedManageDeployment", err)
 	}
-	err = c.ReconcileStackHPA(ctx, sc.Stack, sc.Resources.HPA, sc.GenerateHPA)
+	err = c.ReconcileStackHPA(ctx, sc.Stack, sc.Resources.HPA, sc.GenerateHPA, sc.ScaledDown())
```
linki, line 975: Since both `sc.GenerateHPA` and `sc.ScaledDown()` live on the same object `sc` we might be able to call `ScaleDown` from within `GenerateHPA` which could simplify things.

controller/stack_resources.go
linki: Resolved by moving scale down into GenerateHPA().

PR #538, opened 2023-10-23
Prevent scaling up stacks without traffic

miladbarazandeh:
Remove the HPA when a stack is scaled down and create it again if it scales up, this way we avoid scaling a downscaled stack and other components like kube-metrics-adapter also doesn't need to do work calculating metrics of an HPA which is not actually in use.

https://github.bus.zalan.do/teapot/issues/issues/3482


linki:
👍

gargravarr:
:+1:

mikkeloscar:
Would be nice to have a e2e test for this which shows the HPA is indeed deleted when stack doesn't get traffic.

PR #550, opened 2023-11-02
Revised adaptation of ConfigMap per Stack

katyanna:
**Users provide versioned ConfigMaps explicitly**

```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: my-app
spec:
  stackTemplate:
    spec:
      version: {{{ CDP_BUILD_VERSION }}}
      # Set existing ConfigMap to be owned by Stack
      # Resource must have expected name standard <stackset-name>-<stack-version>-<config-name> 
      configurationResources:
      - configMapRef:
          name: my-app-{{{ CDP_BUILD_VERSION }}}-my-config
      podTemplate:
        spec:
          containers:
          - name: app
            volumes:
            - name: config-volume
              configMap:
                name: my-app-{{{ CDP_BUILD_VERSION }}}-my-config
``` 

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  # Resource must have expected name standard <stackset-name>-<stack-version>-<config-name> 
  name: my-app-{{{ CDP_BUILD_VERSION }}}-my-config
data:
  key1: value1
```


docs/stack_crd.yaml
```diff
@@ -355,14 +355,10 @@ spec:
...
-                      properties:
-                        name:
-                          description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-                            TODO: Add other useful fields. apiVersion, kind, uid?'
-                          type: string
-                      type: object
+                    name:
+                      type: string
+                  required:
+                  - name
```
linki, line 361: Is this schema change necessary? This would require users to change their manifests when we later add the inline solution.

katyanna: That was based on the document, but I agree to keep the `configMapRef` so we can have it for references and later add the `configMap` for inline.

As discussed privately, I'll change it back and update the document.

PR #558, opened 2023-12-14
Add e2e tests for ConfigMap feature

katyanna:
[Debug e2e not failing for ConfigMap feature](https://github.com/zalando-incubator/stackset-controller/pull/557)


linki:
👍

mikkeloscar:
@linki I was think it could make sense to have a test showing the behavior when it fails e.g. the configmap is not there.

Basically traffic switching should not succeed if configmap is missing.

You could be inspired by this test: https://github.com/zalando-incubator/stackset-controller/blob/4278725a69cc7964ac42ac98db2ed7533ab3df33/cmd/e2e/broken_stack_test.go#L13-L53

cmd/e2e/basic_test.go
```diff
@@ -134,6 +144,42 @@ func (f *TestStacksetSpecFactory) Create(stackVersion string) zv1.StackSetSpec {
...
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: configMapName,
+						},
+					},
+				},
+			},
+		}
+	}
```
linki, line 181: This stuff makes sure that if you configure your Stackset to have a versioned ConfigMap (via the factory and calling `ConfigMap()`) that it will end up with a working configuration (name pattern matches, external ConfigMap exists).

linki:
👍

katyanna:
👍

PR #565, opened 2024-01-18
Reference Secrets for per stack versioning

katyanna:
Following the [Reference ConfigMap for per stack versioning](https://github.com/zalando-incubator/stackset-controller/pull/520) and [Revised adaptation of ConfigMap per Stack](https://github.com/zalando-incubator/stackset-controller/pull/550), this Pull Request implements **Secrets versioned by reference**.

The expected behaviour is that given a **Secret** named in the `<stack-name>-<secret-name>` pattern is defined on `stackTemplate.configurationResources.secretRef`, Stackset Controller identifies the deployed **Secret** and updates it with the **Stack** info on `ownerReferences`.

### DIFF
--- a/cmd/e2e/basic_test.go
+++ b/cmd/e2e/basic_test.go
@@ -39,6 +39,7 @@ type TestStacksetSpecFactory struct {
 	routegroup                    bool
 	externalIngress               bool
 	configMap                     bool
+	secret                        bool
 	limit                         int32
 	scaleDownTTL                  int64
 	replicas                      int32
@@ -56,6 +57,7 @@ func NewTestStacksetSpecFactory(stacksetName string) *TestStacksetSpecFactory {
 	return &TestStacksetSpecFactory{
 		stacksetName:           stacksetName,
 		configMap:              false,
+		secret:                 false,
 		ingress:                false,
 		externalIngress:        false,
 		limit:                  4,
@@ -69,6 +71,11 @@ func NewTestStacksetSpecFactory(stacksetName string) *TestStacksetSpecFactory {
 
 func (f *TestStacksetSpecFactory) ConfigMap() *TestStacksetSpecFactory {
 	f.configMap = true
+	return f
+}
+
+func (f *TestStacksetSpecFactory) Secret() *TestStacksetSpecFactory {
+	f.secret = true
 	return f
 }
 
@@ -174,6 +181,40 @@ func (f *TestStacksetSpecFactory) Create(t *testing.T, stackVersion string) zv1.
 						LocalObjectReference: corev1.LocalObjectReference{
 							Name: configMapName,
 						},
+					},
+				},
+			},
+		}
+	}
+
+	if f.secret {
+		secretName := fmt.Sprintf("%s-%s-secret", f.stacksetName, stackVersion)
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: secretName,
+			},
+			Data: map[string][]byte{
+				"key": {0},
+			},
+		}
+
+		_, err := secretInterface().Create(context.Background(), secret, metav1.CreateOptions{})
+		require.NoError(t, err)
+
+		result.StackTemplate.Spec.ConfigurationResources = []zv1.ConfigurationResourcesSpec{
+			{
+				SecretRef: corev1.LocalObjectReference{
+					Name: secretName,
+				},
+			},
+		}
+
+		result.StackTemplate.Spec.PodTemplate.Spec.Volumes = []corev1.Volume{
+			{
+				Name: "secret-volume",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: secretName,
 					},
 				},
 			},
@@ -348,6 +389,22 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 				UID:        stack.UID,
 			},
 		}, configMap.OwnerReferences)
+	}
+
+	// Verify Secrets
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
+		secret, err := waitForSecret(t, rsc.SecretRef.Name)
+		require.NoError(t, err)
+		assert.EqualValues(t, stackResourceLabels, secret.Labels)
+		assert.Contains(t, secret.Name, stack.Name)
+		assert.Equal(t, []metav1.OwnerReference{
+			{
+				APIVersion: "zalando.org/v1",
+				Kind:       "Stack",
+				Name:       stack.Name,
+				UID:        stack.UID,
+			},
+		}, secret.OwnerReferences)
 	}
 
 	verifyStackIngressSources(
@@ -640,6 +697,7 @@ func testStacksetCreate(
 	t *testing.T,
 	testName string,
 	configmap bool,
+	secret bool,
 	hpa,
 	ingress,
 	routegroup,
@@ -655,6 +713,9 @@ func testStacksetCreate(
 		stacksetSpecFactory := NewTestStacksetSpecFactory(stacksetName)
 		if configmap {
 			stacksetSpecFactory.ConfigMap()
+		}
+		if secret {
+			stacksetSpecFactory.Secret()
 		}
 		if hpa {
 			stacksetSpecFactory.Autoscaler(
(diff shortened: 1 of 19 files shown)


cmd/e2e/basic_test.go
```diff
@@ -180,6 +187,40 @@ func (f *TestStacksetSpecFactory) Create(t *testing.T, stackVersion string) zv1.
...
 	}
 
+	if f.secret {
+		secretName := fmt.Sprintf("%s-%s-secret", f.stacksetName, stackVersion)
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: secretName,
+			},
+			Data: map[string][]byte{
+				"key": {0},
```
linki, line 197: Instead of `{0}` we can also use `[]byte("value")` so that we have a more readable value.

cmd/e2e/basic_test.go
```diff
@@ -350,6 +391,22 @@ func verifyStack(t *testing.T, stacksetName, currentVersion string, stacksetSpec
 		}, configMap.OwnerReferences)
 	}
 
+	// Verify Secrets
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
+		secret, err := waitForSecret(t, rsc.SecretRef.Name)
+		require.NoError(t, err)
+		assert.EqualValues(t, stackResourceLabels, secret.Labels)
+		assert.Contains(t, secret.Name, stack.Name)
```
linki, line 399: Let's also validate the _value_ for completeness, similar to what we do for ConfigMaps (`assert.Equal(t, map[string]string{"key": "value"}, configMap.Data)`)

controller/stack_resources.go
```diff
@@ -54,6 +54,25 @@ func equalResourceList(
 	return reflect.DeepEqual(existingName, crName)
 }
 
+func equalSecretList(
```
linki, line 57: This function is practically identical to `equalConfigMapList` (only difference is one takes Secrets and one takes ConfigMap and both have the .Name field). We should remove as much duplication as possible here.

katyanna: Yes, I'm running in draft to check the e2e but the PR is still going to be refactored.

controller/stack_resources.go
```diff
@@ -397,3 +418,62 @@ func (c *StackSetController) ReconcileStackConfigMap(
 	}
 	return nil
 }
+
+func (c *StackSetController) ReconcileStackSecret(
```
linki, line 422: This function is very similar to `ReconcileStackConfigMap`. Maybe it's worth checking whether a function like `ReconcileStackConfigurationResources()` where we iterate over `stack.Spec.ConfigurationResources` once makes sense.

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+					SecretRef: v1.LocalObjectReference{
+						Name: "foo-v1-scnd-secret",
+					},
+				},
+			},
+		},
+	}
+
+	baseData := map[string][]byte{
+		"testK": {1, 2},
```
linki, line 1250: `[]byte("some string")` also works I think and allows for more readable values.

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+			name:     "secret ownerReference is added to referenced secret",
+			stack:    testSecretStack,
+			existing: nil,
+			template: []*v1.Secret{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "foo-v1-test-secret",
+						Namespace: testSecretStack.Namespace,
+					},
+
```
linki, line 1301: nit: extra line here

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+				{
+					SecretRef: v1.LocalObjectReference{
+						Name: "foo-v1-test-secret",
+					},
+				},
+			},
+		},
+	}
+
+	testMixedConfigtStack := baseTestStack
```
linki, line 1213: typo: testMixedConfig**t**Stack

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+			name:     "manage secrets out of mixed configurationResources",
+			stack:    testMixedConfigtStack,
+			existing: nil,
+			template: []*v1.Secret{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "foo-v1-test-secret",
+						Namespace: testSecretStack.Namespace,
+					},
+
```
linki, line 1378: nit: extra line here

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+					err = env.CreateSecrets(context.Background(), []v1.Secret{*i})
+					require.NoError(t, err)
+				}
+			}
+
+			if tc.existing != nil {
+				for _, existing := range tc.existing {
+					err = env.CreateSecrets(context.Background(), []v1.Secret{*existing})
+					for _, template := range tc.template {
+						if existing.Name == template.Name {
```
linki, line 1407: I don't really understand this test.

For each to-be-existing secret, we try to create it. Then we check each template and assert that the creation raised an error if the name of the secret matches the template name (and vice versa).

controller/stack_resources_test.go
```diff
@@ -1195,3 +1195,241 @@ func TestReconcileStackConfigMap(t *testing.T) {
...
+					}
+				}
+			}
+
+			err = env.controller.ReconcileStackSecret(
+				context.Background(), &tc.stack, tc.existing, func(tmp *metav1.ObjectMeta) *metav1.ObjectMeta {
+					if tmp.Name == tc.template[0].Name {
+						return &tc.expected[0].ObjectMeta
+					}
+					return &tc.expected[1].ObjectMeta
```
linki, line 1421: Looks like we're mocking the response of `sc.UpdateObjectMeta` here and then return either `tc.expected[0]` or `tc.expected[1]` based on the `tmp.Name`. Can we do something like the following instead?

Make `tc.expected` a `map[string]v1.Secret` instead of a `[]*v1.Secret` and then use the `tmp.Name` as the key to get the right secret:
```go
func(tmp *metav1.ObjectMeta) *metav1.ObjectMeta {
  return &tc.expected[tmp.Name].ObjectMeta
}
```

linki: Maybe we shouldn't mock it in the first place and rather work with the real implementation. It's not that it's doing that much that needs mocking (such as network calls and so on).

https://github.com/zalando-incubator/stackset-controller/blob/d0f7901c3b4101dec8fdd7289be8525e0235d099/pkg/core/stack_resources.go#L536-L543

katyanna: I mock it to isolate the test. Since this test is about the `ReconcileStackSecrets` specifically, I'd rather not have other functions interfering with it. `UpdateObjectMeta` has its own test also.

controller/stack_resources.go
katyanna: > where we iterate over `stack.Spec.ConfigurationResources` once and handle each item based on its type.

That's on the 4th line already. They're similar, but the type constraint starts in the parameters. I also think it makes sense to handle one resource kind per function, as happens in the rest of the code, specially considering that the `ReconcileStackConfigMap` will be expanded for the inline option.

controller/stack_resources.go
katyanna: I couldn't find an elegant solution to merge the functions, as they differ in types. Keeping the main function concise and with clear purpose is of more importance so I'd rather not pollute it with type management for a helper function. But let me know if you have an idea about how to make it more DRY.

controller/stackset.go
```diff
@@ -1459,15 +1498,23 @@ func resourceReadyTime(timestamp time.Time, ttl time.Duration) bool {
 	return false
 }
 
-// validateConfigurationResourceNames returns an error if any ConfigurationResource name is not prefixed by Stack name.
+// validateConfigurationResourceNames returns an error if any ConfigurationResource
+// name is not prefixed by Stack name.
 func validateConfigurationResourceNames(stack *zv1.Stack) error {
+	var rscName string
```
mikkeloscar, line 1504: This should be scoped to the for loop e.g. move to line 1505

controller/stack_resources_test.go
katyanna: This is setup for the actual tests. It creates the `existing` Secrets of the test case and errors if there's a template in that test case with the same name. May be best to have it in opposite orders - first create the existing and when creating the template, check is the name is already there.

controller/stack_resources.go
```diff
@@ -397,3 +420,74 @@ func (c *StackSetController) ReconcileStackConfigMap(
...
+		return nil
+	}
+
+	for _, rsc := range stack.Spec.ConfigurationResources {
+		if rsc.SecretRef.Name == "" {
+			continue
+		}
+		rscName := rsc.SecretRef.Name
+
+		if err := validateConfigurationResourcesNames(stack); err != nil {
```
linki, line 456: This function validates _all_ `ConfigurationResources` so it needs to be _outside_ loop (or changed).

controller/stack_resources.go
```diff
@@ -397,3 +420,74 @@ func (c *StackSetController) ReconcileStackConfigMap(
...
+	ctx context.Context,
+	stack *zv1.Stack,
+	existing []*apiv1.Secret,
+	updateObjMeta func(*metav1.ObjectMeta) *metav1.ObjectMeta,
+) error {
+	if stack.Spec.ConfigurationResources == nil {
+		return nil
+	}
+
+	if equalSecretList(existing, stack.Spec.ConfigurationResources) {
```
linki, line 446: How is that even correct?

This line says that when _all_ of the `ConfigurationResources` already exist (`existing`) we return early and don't to any processing. But in our _revised_ solution we don't create any Secrets anymore (the user's provide them), therefore the Secrets exist from the very beginning. I wonder how we'll ever get past this point. Of course there could be certain race conditions where the `.configurationResources` section is created at a different time than the referenced Secret but I don't think we want to rely on that.

katyanna: Yes, these checks are no longer needed. Thanks for the observation.

e2e/apply/rbac.yaml
```diff
@@ -90,6 +90,17 @@ rules:
...
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
```
linki, line 103: nit: We could reduce this set of the permissions. We don't need to delete and create (same for ConfigMaps)

katyanna: We still them for the e2e.

linki: Ok, that makes sense 😄 🤦

controller/stack_resources.go
```diff
@@ -351,15 +328,13 @@ func (c *StackSetController) ReconcileStackConfigMap(
 		return nil
```
linki, line 328: This `if` should also be removable now since the function only contains the `range` call (which works fine with `nil`)

controller/stackset.go
```diff
@@ -1459,15 +1500,31 @@ func resourceReadyTime(timestamp time.Time, ttl time.Duration) bool {
...
+		var rscName string
+		if rsc.ConfigMapRef.Name != "" {
+			rscName = rsc.ConfigMapRef.Name
+		}
+
+		if rsc.SecretRef.Name != "" {
+			rscName = rsc.SecretRef.Name
+		}
+
 		if !strings.HasPrefix(rscName, stack.Name) {
```
linki, line 1516: This can reuse the `validateConfigurationResourceName` function created below.

katyanna: Tried that, it looks awkward and generates a go warning. I think we don't need to get so strict here.

linki: This should've done it.

```go
		if err := validateConfigurationResourceName(stack.Name, rscName); err != nil {
			return err
		}
```

The point is to not have the `HasPrefix` check twice because it's very easy to forget to change both in case the logic ever changes.

linki: Let's do some refactorings in other PRs.

linki:
👍

controller/stack_resources.go
linki: That's fine but we can still reduce the duplication, e.g. by extracting some code into helper functions.

katyanna:
👍

PR #566, opened 2024-01-25
introduce namespace cli flag to limit scope

linki:
Adds a `--namespace` flag to optionally configure a single namespace in cases where you want to avoid global acccess.


mikkeloscar:
:+1:

PR #567, opened 2024-01-30
Isolate permissions of e2e stackset-controller to its own namespace to avoid clashes

linki:
Since introducing the [`--namespace` flag](https://github.com/zalando-incubator/stackset-controller/pull/566) and running e2e tests with it, we can reduce stackset-controller's permissions to a single namespace as well.

This is less subtle than it seems to be as this allows for _correctly_ running multiple stackset-controller e2e test runs in parallel (i.e. in different concurrent PRs). We did this before but it was flawed because some resources, such as the [`ClusterRole`](https://github.com/zalando-incubator/stackset-controller/blob/c9410526ba568e2ec8e9fc7bddad8895ceff4f4b/e2e/apply/rbac.yaml#L8-L10), were shared between different open PRs and therefore would overwrite each other, leading to unpredictable results when changed.

By isolating every e2e run into a separate namespace, we'll avoid running into any of these problems.

### DIFF
--- a/e2e/apply/rbac.yaml
+++ b/e2e/apply/rbac.yaml
@@ -5,7 +5,7 @@ metadata:
   name: stackset-controller
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
   name: stackset-controller
 rules:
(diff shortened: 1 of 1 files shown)


katyanna:
👍

PR #575, opened 2024-02-01
Enable traffic segments on annotated StackSets only with --annotated-traffic-segments.

gargravarr:
Right now to enable traffic segments via annotation we need to set both `--enable-traffic-segments` and `--annotated-traffic-segments`. This can be dangerous, because it is very easy to set first `--enable-traffic-segments` which would trigger migration to traffic segments of _all_ StackSets.

With this Pull Request you will only need to set `--annotated-traffic-segments` to enable traffic segments via annotations. I this way, we only need to set this flag in the corresponding migration stages.

Additionally, this PR formats the code with `goimports`.

### DIFF
--- a/controller/stackset.go
+++ b/controller/stackset.go
@@ -249,11 +249,11 @@ func (c *StackSetController) injectSegmentAnnotation(
 	ctx context.Context,
 	stackSet *zv1.StackSet,
 ) bool {
-	if !c.trafficSegmentsEnabled {
+	if c.annotatedTrafficSegments {
 		return false
 	}
 
-	if c.annotatedTrafficSegments {
+	if !c.trafficSegmentsEnabled {
 		return false
 	}
 
@@ -310,13 +310,14 @@ func (c *StackSetController) collectResources(ctx context.Context) (map[types.UI
 		}
 
 		stacksetContainer := core.NewContainer(&stackset, reconciler, c.backendWeightsAnnotationKey, c.clusterDomains)
-		if c.trafficSegmentsEnabled &&
-			stackset.Annotations[TrafficSegmentsAnnotationKey] == "true" {
+		if c.trafficSegmentsEnabled || c.annotatedTrafficSegments {
 
-			stacksetContainer.EnableSegmentTraffic()
-			stacksetContainer.SynchronizeIngressAnnotations(
-				c.syncIngressAnnotations,
-			)
+			if stackset.Annotations[TrafficSegmentsAnnotationKey] == "true" {
+				stacksetContainer.EnableSegmentTraffic()
+				stacksetContainer.SynchronizeIngressAnnotations(
+					c.syncIngressAnnotations,
+				)
+			}
 		}
 		stacksets[uid] = stacksetContainer
 	}
--- a/controller/stackset_test.go
+++ b/controller/stackset_test.go
@@ -328,7 +328,9 @@ func TestSegmentEnablement(t *testing.T) {
 				testStackset("foo", "default", "123"),
 				testStacksetAnnotatedSegment("bar", "default", "456"),
 			},
-			segmentsEnabled: map[types.UID]bool{"123": false, "456": false},
+			annotatedTrafficSegments: false,
+			trafficSegmentsEnabled:   false,
+			segmentsEnabled:          map[types.UID]bool{"123": false, "456": false},
 		},
 		{
 			stacksets: []zv1.StackSet{
@@ -336,6 +338,7 @@ func TestSegmentEnablement(t *testing.T) {
 				testStacksetAnnotatedSegment("bar", "default", "0AB"),
 			},
 			annotatedTrafficSegments: true,
+			trafficSegmentsEnabled:   false,
 			segmentsEnabled:          map[types.UID]bool{"789": false, "0AB": true},
 		},
 		{
@@ -343,8 +346,9 @@ func TestSegmentEnablement(t *testing.T) {
 				testStackset("foo", "default", "CDE"),
 				testStacksetAnnotatedSegment("bar", "default", "FGH"),
 			},
-			trafficSegmentsEnabled: true,
-			segmentsEnabled:        map[types.UID]bool{"CDE": false, "FGH": true},
+			annotatedTrafficSegments: false,
+			trafficSegmentsEnabled:   true,
+			segmentsEnabled:          map[types.UID]bool{"CDE": false, "FGH": true},
 		},
 		{
 			stacksets: []zv1.StackSet{
--- a/controller/stackset_test.go
+++ b/controller/stackset_test.go
@@ -318,10 +318,10 @@ func TestCollectResources(t *testing.T) {
 
 func TestSegmentEnablement(t *testing.T) {
 	for _, tc := range []struct {
-		stacksets []zv1.StackSet
+		stacksets                []zv1.StackSet
 		annotatedTrafficSegments bool
-		trafficSegmentsEnabled bool
-		segmentsEnabled map[types.UID]bool
+		trafficSegmentsEnabled   bool
+		segmentsEnabled          map[types.UID]bool
 	}{
 		{
 			stacksets: []zv1.StackSet{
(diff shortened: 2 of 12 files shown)


gargravarr:
:+1:

controller/stackset_test.go
```diff
@@ -316,6 +316,79 @@ func TestCollectResources(t *testing.T) {
...
+			annotatedTrafficSegments: true,
+			segmentsEnabled: map[types.UID]bool{"789": false, "0AB": true},
+		},
+		{
+			stacksets: []zv1.StackSet{
+				testStackset("foo", "default", "CDE"),
+				testStacksetAnnotatedSegment("bar", "default", "FGH"),
+			},
+			trafficSegmentsEnabled: true,
+			segmentsEnabled: map[types.UID]bool{"CDE": false, "FGH": true},
```
linki: What is the difference to the test before? Does the `CDE` vs. `789` mean anything?

controller/stackset_test.go
```diff
@@ -316,6 +316,79 @@ func TestCollectResources(t *testing.T) {
...
+		},
+		{
+			stacksets: []zv1.StackSet{
+				testStackset("foo", "default", "IJK"),
+				testStacksetAnnotatedSegment("bar", "default", "LMN"),
+			},
+			annotatedTrafficSegments: true,
+			trafficSegmentsEnabled: true,
+			segmentsEnabled: map[types.UID]bool{"IJK": false, "LMN": true},
```
linki: Why is `IJK: false` when we enable TrafficSegment everywhere? (at least it looks like the test is for that case)

controller/stackset_test.go
```diff
@@ -316,6 +316,79 @@ func TestCollectResources(t *testing.T) {
...
+		},
```
linki, line 357: Let's also add a test case for?

```
			annotatedTrafficSegments: false,
			trafficSegmentsEnabled: true,
```

controller/stackset_test.go
linki: I see the difference now, nevermind.

linki: Shouldn't this _include_ non-annotated stacksets? `CBE`

gargravarr: The difference is that the test before only has `annotatedTrafficSegments` set, while this test only has `trafficSegmentsEnabled` set.

`CDE` and `789` are the UIDS of the StackSets which the stackset controller references.

controller/stackset_test.go
gargravarr: This test has `annotatedTrafficSegments` set to `false` (zero value for `bool`): https://github.com/zalando-incubator/stackset-controller/blob/safer-flags/controller/stackset_test.go#L341-L348

controller/stackset_test.go
gargravarr: It does include. Line 344, StackSet with UID `FGH`

controller/stackset_test.go
gargravarr: Because StackSet `IJK` does not contain the annotation _yet_. The StackSet controller will inject our special annotation on _all_ StackSets when `TrafficSegmentsEnabled` is `true`, here is the code: https://github.com/zalando-incubator/stackset-controller/blob/safer-flags/controller/stackset.go#L248

This is so that deployment-service and cdp-steps can distinguish which StackSets have/don't have segments.

controller/stackset_test.go
linki: How about making that more apparent? Like specifically setting both variables to either true/false in each of the four test cases? It would be easier to notice that all possible combinations are checked.

controller/stackset_test.go
linki: Right, but I mean `CBE`. It's `false` although  `trafficSegmentsEnabled: true` (traffic segments enabled for _all_ stacksets). I'm probably missing something, but it's fine if you say it's correct! 👍

controller/stackset.go
```diff
@@ -310,13 +310,14 @@ func (c *StackSetController) collectResources(ctx context.Context) (map[types.UI
...
-			stackset.Annotations[TrafficSegmentsAnnotationKey] == "true" {
+		if c.trafficSegmentsEnabled || c.annotatedTrafficSegments {
 
-			stacksetContainer.EnableSegmentTraffic()
-			stacksetContainer.SynchronizeIngressAnnotations(
-				c.syncIngressAnnotations,
-			)
+			if stackset.Annotations[TrafficSegmentsAnnotationKey] == "true" {
+				stacksetContainer.EnableSegmentTraffic()
+				stacksetContainer.SynchronizeIngressAnnotations(
```
linki, line 317: Does that mean that TrafficSegments are disabled and annotations are not synced for non-annotated Stacksets?

gargravarr: Yes. As I explained, when we enable segments by default, the StackSet controller will inject the annotation on all StackSets, here is the code: https://github.com/zalando-incubator/stackset-controller/blob/safer-flags/controller/stackset.go#L248

We need this annotation so that cdp-steps and deployments-service also work.

controller/stackset_test.go
gargravarr: Good idea, I will do that.

controller/stackset_test.go
gargravarr: So that I won't repeat myself, and I agree it's not straightforward to understand how segments by default work, it's the same explanation as [here](https://github.com/zalando-incubator/stackset-controller/pull/575#discussion_r1475782878).

controller/stackset_test.go
gargravarr: Done!

controller/stackset_test.go
```diff
@@ -316,6 +316,83 @@ func TestCollectResources(t *testing.T) {
...
 
+func TestSegmentEnablement(t *testing.T) {
+	for _, tc := range []struct {
+		stacksets                []zv1.StackSet
+		annotatedTrafficSegments bool
+		trafficSegmentsEnabled   bool
+		segmentsEnabled          map[types.UID]bool
+	}{
+		{
+			stacksets: []zv1.StackSet{
```
mikkeloscar, line 327: It would be good with a test title to describe the intention of the test.

It's common to store this in a field `msg` or similar and then pass it like this for running the tests:

```go
t.Run(tc.msg, func(t testing.T) {
```

Then you also see it in the go test output more clearly :)

controller/stackset_test.go
```diff
@@ -316,6 +316,83 @@ func TestCollectResources(t *testing.T) {
...
+			segmentsEnabled:          map[types.UID]bool{"789": false, "0AB": true},
+		},
+		{
+			stacksets: []zv1.StackSet{
+				testStackset("foo", "default", "CDE"),
+				testStacksetAnnotatedSegment("bar", "default", "FGH"),
+			},
+			annotatedTrafficSegments: false,
+			trafficSegmentsEnabled:   true,
+			segmentsEnabled:          map[types.UID]bool{"CDE": false, "FGH": true},
```
mikkeloscar, line 351: Should this not be `true` for both? :thinking:

gargravarr: No. Stack `FGH` has the annotation.

controller/stackset_test.go
gargravarr: I really find that annoying, because I just found silly that on a _tesing_ funciton we just call the upward `Run` function.

But this test is really not straightfoward to read, so I'll update it to use `t.Run`, with a grin on my face.

gargravarr: Done!

controller/stackset_test.go
mikkeloscar: I don't understand why setting `trafficSegmentsEnabled` would not apply to all stacks regardless of the annotation.

I guess it's the detail of the "next loop" meaning it would be true if run twice?

controller/stackset_test.go
mikkeloscar: Thanks!

> that on a testing function we just call the upward Run function.

But the passed `t testing.T` to the function you define is the way the Go test framework makes the features of the framework available. So it's not that `Run` is an upward function, but rather a function intended for exactly this use case where you want to run more tests with the context of a test title. The `func(t testing.T) {` is just the way or propagating the framework features to the next level so you can e.g fail the test.

mikkeloscar:
:+1:

controller/stackset_test.go
gargravarr: Exactly, as I explained [here](https://github.com/zalando-incubator/stackset-controller/pull/575#discussion_r1475782878).

gargravarr:
:+1:

PR #577, opened 2024-02-05
extract processing of one configuration resource into separate function

linki:
This more of less only moves inner loop into a separate function. Little easier to think about and maybe opens up some opportunity to combine the `ReconcileStackConfigMap` and `ReconcileStackSecret` functions since they are very similar.


katyanna:
:+1:

controller/stack_resources.go
```diff
@@ -380,10 +406,9 @@ func (c *StackSetController) ReconcileStackConfigMap(
 // running resources is also allowed, so the method checks for changes on the
 // ConfigurationResources to ensure all listed Secrets are properly linked
 // to the Stack.
-func (c *StackSetController) ReconcileStackSecret(
+func (c *StackSetController) ReconcileStackSecrets(
 	ctx context.Context,
 	stack *zv1.Stack,
-	existing []*apiv1.Secret,
```
linki, line 386: @katyanna What I find interesting: If we don't need, i.e. care about, the existing resources, we probable don't need to run [collectConfigMaps](https://github.com/zalando-incubator/stackset-controller/blob/1d30fdbb78b97a47b421bb6b5a2f97d773df6c6e/controller/stackset.go#L564) either 😄 

But I think we'll need it for the inline solution, so let's just keep it.

controller/stack_resources.go
```diff
@@ -318,10 +328,9 @@ func (c *StackSetController) ReconcileStackRouteGroup(ctx context.Context, stack
 // running resources is also allowed, so the method checks for changes on the
 // ConfigurationResources to ensure all listed ConfigMaps are properly linked
 // to the Stack.
-func (c *StackSetController) ReconcileStackConfigMap(
+func (c *StackSetController) ReconcileStackConfigMaps(
 	ctx context.Context,
 	stack *zv1.Stack,
-	existing []*apiv1.ConfigMap,
```
linki, line 324: @katyanna What I find interesting: If we don't need, i.e. care about, the existing resources, we probable don't need to run [collectConfigMaps](https://github.com/zalando-incubator/stackset-controller/blob/1d30fdbb78b97a47b421bb6b5a2f97d773df6c6e/controller/stackset.go#L564) either 😄 

But I think we'll need it for the inline solution, so let's just keep it.

katyanna:
:+1:

linki:
👍

katyanna:
👍

PR #578, opened 2024-02-05
extract owner detection of configuration resources into separate function

linki:
The owner detection loop looks a bit complicated. Extracting some of it into a shared function.

It's not much less code but due to using the named intermediate variables one can reason about what's happening: "Raise error if the resource is owned unless it's us in which case we just return. If it's not already owned, continue processing it."


katyanna:
:+1:

linki:
👍

PR #579, opened 2024-02-05
refactor TestReconcileStackConfigMaps

linki:
This refactors the TestReconcileStackConfigMaps and renames the variables a bit.

Main change though is how the `template` are added to the stack. Before these were Configmaps that were created as part of the test and needed some complicated error logic.

image

Instead they were moved into the `existing` (renamed to `existingConfigMaps`) field because they are created before the test start. Looking at the old version of the test there was no difference between the `existing` and the `template` fields, both were just created via `CreateConfigMaps`.

Furthermore, a new field was introduced containing the ConfigurationResourceReferences that the Stack-under-test should have (`configurationResources`). This avoids the need to have a separate Stack to test the multiple-ConfigResources cases amongst other things, such as makes the test cases more self-contained. It still required the need to extract a lot of variables outside of each test case (before the `tc` loop) to avoid all the boilerplate code.

I still don't agree with mocking `updateObjMeta` because it's kind of the _heart_ of this test case (turning a ConfigMap without `ownerReferences` into one that has them) and exactly this part is completely mocked out. Nevertheless it tests the other parts of the function, iterating over the referenced configurationResources, finding them etc.


linki:
👍

controller/stack_resources.go
```diff
@@ -31,6 +32,15 @@ func syncObjectMeta(target, source metav1.Object) {
 	target.SetAnnotations(source.GetAnnotations())
 }
 
+// isOwned checks if the resource is owned and returns the UID of the owner.
+func isOwned(ownerReferences []metav1.OwnerReference) (bool, types.UID) {
```
gargravarr, line 36: do we really need this function? Isn't `len(ownerReferences) != 0` enough?

controller/stack_resources.go
```diff
@@ -333,42 +342,59 @@ func (c *StackSetController) ReconcileStackConfigMapRefs(
...
+	if err != nil {
+		return err
+	}
 
-		_, err = c.client.CoreV1().ConfigMaps(configMap.Namespace).
-			Update(ctx, configMap, metav1.UpdateOptions{})
-		if err != nil {
-			return err
+	// Check if the ConfigMap is already owned by us or another resource.
+	// If the ConfigMap is owned by another resource, we should not update it.
```
gargravarr, line 366: It seems also that the function also won't update the ConfigMap also if it owns it, just won't return an error. Am I missing something?

controller/stack_resources.go
```diff
@@ -395,37 +420,53 @@ func (c *StackSetController) ReconcileStackSecretRefs(
...
+	if err != nil {
+		return err
+	}
 
-		_, err = c.client.CoreV1().Secrets(secret.Namespace).
-			Update(ctx, secret, metav1.UpdateOptions{})
-		if err != nil {
-			return err
+	// Check if the Secret is already owned by us or another resource.
+	// If the Secret is owned by another resource, we should not update it.
```
gargravarr, line 443: Similar question as [here](https://github.com/zalando-incubator/stackset-controller/pull/579/files?diff=unified&w=1#r1507307150).

controller/stack_resources.go
linki: It's doing more than that. It also returns the found owner, if any.

So it usually becomes as cumbersome as:
```go
if len(ownerReferences) != 0 {
  // do something with ownerReferences[0].UID
}
```

And this could become more complicated when we actually process the list of owners instead of just taking the first one.

With this we can just hide this weird concept of multiple owners a little bit.

linki: Of course, we can simplify it to `getOwner()` that can return an empty string instead of an additional bool.

Then any check for the more existence of an owner would be come `getOwner() != ""` which is still pretty good.

linki: For additional context: My goal would be to extract certain often-used functions into our shared Go library at some point. Getting the owner reliably and consistent is one of those functions because we do that in all kinds of places _and_ use different approaches sometimes.

For this function to be extracted and used throughout several projects, I think it must be a little bit more "elaborate" to be useful across multiple projects without being too complicated again. In this example that means having the `bool` for mere owner existence checks so that I can do a simple `ok, _ := isOwner()` if I only care about that part.

gargravarr: Ah, now that you bring the use in a conditional, yes it's more practical as a function.

controller/stack_resources.go
gargravarr: So the update would just happen only if the resource is orphaned? If so a comment would make this more clear, yes.

linki: If I just [revert this commit](https://github.com/zalando-incubator/stackset-controller/pull/579/commits/560773617497980f4716419c0f2680759f47100f) would that solve your concern?

gargravarr: Looks more clear, yes.

linki: Ok, done

controller/stack_resources.go
linki: Resolved, see other comment thread.

linki:
👍

demonCoder95:
👍

mikkeloscar:
:+1:

PR #580, opened 2024-02-08
Add PlatformCredentialsSet API client 

katyanna:
Preparing for the implementation of the **PlatformCredentialsSet per Stack** feature,
this Pull Request adds `PlatformCredentialsSet` to the zalando.org/v1 api.


mikkeloscar:
@katyanna I think you also need to add the new types here: https://github.com/zalando-incubator/stackset-controller/blob/2104a9601340baee5267a02040d98c8d4820cdd8/pkg/apis/zalando.org/v1/register.go#L35

katyanna:
:+1:

pkg/apis/zalando.org/v1/types_pcs.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// PlatformCredentialsStatus is the status part of the Stack.
+// +k8s:deepcopy-gen=true
+type PlatformCredentialsStatus struct {
+	ObservedGeneration int64               `json:"observedGeneration"`
+	Errors             []string            `json:"errors"`
+	Problems           []string            `json:"problems"`
+	Tokens             map[string]struct{} `json:"tokens"`
+	Clients            map[string]struct{} `json:"clients"`
```
mikkeloscar, line 48: Ideally we should have proper types defined here, but it's probably not relevant for the implementation in stackset.

mikkeloscar:
:+1:

pkg/apis/zalando.org/v1/types_pcs.go
AlexanderYastrebov: This actually fails with
```console
/masked/github.com/zalando-incubator/stackset-controller/pkg/apis/zalando.org/v1/types_pcs.go:47:32: not a supported map value type: *ast.StructType
/masked/github.com/zalando-incubator/stackset-controller/pkg/apis/zalando.org/v1/types_pcs.go:48:32: not a supported map value type: *ast.StructType
Error: not all generators ran successfully
```
but the failure is masked by `|| /bin/true || true`:
https://github.com/zalando-incubator/stackset-controller/blob/2bcb26cc48580356857eae40e1399b5893a90a26/Makefile#L38

introduced ages ago by #182

Maybe it was necessary back then but now I think it is not, I tried to clean build a commit before this one (2104a96) with updated Makefile (used old go version due to #610)
```diff
diff --git a/Makefile b/Makefile
index ecd5ca5..5e4ea29 100644
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ $(GENERATED): go.mod $(CRD_TYPE_SOURCE)
        ./hack/update-codegen.sh
 
 $(GENERATED_CRDS): $(GENERATED) $(CRD_SOURCES)
-       go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:crdVersions=v1,allowDangerousTypes=true paths=./pkg/apis/... output:crd:dir=docs || /bin/true || true
+       ~/bin/go1.21.5/bin/go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:crdVersions=v1,allowDangerousTypes=true paths=./pkg/apis/... output:crd:dir=docs
```
and it seems to work (not sure why it adds `x-kubernetes-map-type: atomic` everywhere though, maybe due to https://github.com/kubernetes-sigs/controller-tools/pull/753)

mikkeloscar: Here is an improvement of this: https://github.com/zalando-incubator/stackset-controller/pull/614

PR #584, opened 2024-02-13
Ensure controller only deletes central Ingress/RouteGroup when all segments exist.

gargravarr:
This Pull Request adds a safety check when converting StackSet resources from central ingress to ingess segments: with this check the controller will only delete the central Ingress (or RouteGroup) if and only if all Stacks defining an Ingress and/or a RouteGroup also contain theis segment counterparts.


linki:
👍

gargravarr:
:+1:

PR #586, opened 2024-02-15
make local object reference a pointer so it can really be optional

linki:
The optional subtypes of each configurationResource will always be rendered. This works file but looks weird, especially with more types. Let's see if making it a pointer fixes it.

image

There's no `secretRef` in the YAML but the in-cluster object has an empty struct.


mikkeloscar:
:+1:

katyanna:
👍

PR #587, opened 2024-02-15
Support PlatformCredentialsSet versioning

katyanna:
```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: {{STACKSET_NAME}}
spec:
  {...}
  stackTemplate:
    version: {{VERSION}}
    spec:
      configurationResources:
        - platformCredentialsSet:
             name: {{STACKSET_NAME}}-{{VERSION}}-my-credentials
             tokens:
               sample-token:
                 privileges:
                 - com.zalando::foobar.write
```

From this definition, the Stackset controller creates a **PlatformCrendentialsSet**
named `pcs-dev-1-my-credentials` owned by the **Stack**.


katyanna:
todo:
- [ ] Fix update (check generation annotation)
- [ ] Add unit tests
- [ ] Add e2e tests
- [ ] Add doc comments

docs/stack_crd.yaml
```diff
@@ -348,9 +348,8 @@ spec:
                 - metrics
                 type: object
               configurationResources:
-                description: ConfigurationResources describes the ConfigMaps that
-                  will be created. Later Secrets and PlatformCredentialSets will also
-                  be defined on ConfigurationResources
+                description: ConfigurationResources describes the ConfigMaps, Secrets,
+                  and/or PlatformCresentialsSet that will be created.
```
linki, line 352: typo: PlatformCre*s*entialsSet

docs/stack_crd.yaml
```diff
@@ -364,6 +363,78 @@ spec:
...
+                                      type: string
+                                    type: array
+                                required:
+                                - privileges
+                                type: object
+                              type: object
+                          required:
+                          - application
+                          type: object
+                        status:
```
linki, line 408: The PCS's `status` section is probably not needed here.

docs/stackset_crd.yaml
```diff
@@ -621,9 +621,8 @@ spec:
                         - metrics
                         type: object
                       configurationResources:
-                        description: ConfigurationResources describes the ConfigMaps
-                          that will be created. Later Secrets and PlatformCredentialSets
-                          will also be defined on ConfigurationResources
+                        description: ConfigurationResources describes the ConfigMaps,
+                          Secrets, and/or PlatformCresentialsSet that will be created.
```
linki, line 625: The same comments as for `docs/stack_crd.yaml` apply here, too.

docs/zalando.org_platformcredentialssets.yaml
```diff
@@ -32,26 +32,13 @@ spec:
             type: string
           metadata:
             type: object
+          name:
+            type: string
```
linki, line 36: I don't remember the purpose of this file. It looks like the definition of the PCS CRD that can be applied to the cluster. In this case we should keep it close to [its real definition](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/01-platformcredentialsset/customresourcedefinition.yaml) if not the same.

linki:
What happened to the alternative definition described [in the doc](https://docs.google.com/document/d/1qboqAKuqfUOx7VZr_V9SFq5n6r5yBVXibRTZFapvCPY/edit#heading=h.97gevkjcy1sr)where there's one level less of YAML and the application is detected from the StackSet, e.g.

```yaml
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: pcs
  labels:
    application: pcs-app
spec:
  {...}
  stackTemplate:
    version: "dev-1"
    spec:
      configurationResources:
        - platformCredentialsSet:
            name: my-credentials
            tokens:
              sample-token:
                privileges:
                - com.zalando::foobar.write
```

docs/zalando.org_platformcredentialssets.yaml
linki: If this is our own local definition to access the PCS it doesn't need to be the same. In this case we can clean it up further, remove `apiVersion`, `kind`, etc.

controller/stackset.go
```diff
@@ -590,6 +603,33 @@ func (c *StackSetController) collectSecrets(ctx context.Context, stacksets map[t
...
 }
 
+func (c *StackSetController) collectPlatformCredentialsSet(
+	ctx context.Context,
+	stacksets map[types.UID]*core.StackSetContainer,
+) error {
+	platformCredentialsSets, err := c.client.ZalandoV1().PlatformCredentialsSets(c.namespace).
+		List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list platformCredentialsSet: %v", err)
```
linki, line 613: Let's upper-case this to keep it aligned with the others: `failed to list PlatformCredentialsSet`

docs/zalando.org_platformcredentialssets.yaml
mikkeloscar: It doesn't really have a purpose here, we just automatically generate it when adding the PCS types in `types.go` we could also exclude it, but it's maybe helpful to have to compare to the real definition to see if the types.go is defined correctly :)

controller/stack_resources.go
```diff
@@ -429,3 +429,77 @@ func (c *StackSetController) ReconcileStackSecret(
...
+	existing []*zv1.PlatformCredentialsSet,
+	generateUpdated func(zv1.ConfigurationResourcesSpec) (*zv1.PlatformCredentialsSet, error),
+) error {
+	pcs, err := generateUpdated(rsc)
+	if err != nil {
+		return err
+	}
+
+	// Create new PlatformCredentialsSet
+	if existing == nil {
```
linki, line 466: `existing` is an array of existing `PlatformCredentialsSets` (whereas it's just a single object in the similar functions above). That means `nil` means that there are no PCS _at all_ for the StackSet (hence we're going to create the PCS, so far so good). However, when there's at least one PCS present, we won't create any new ones.

This usually doesn't happen, I think, because we create all PCS _at once_ and if we ever add/remove a PCS from the list we're going to deploy a new Stack anyways. However, with the current implementation it can get in an inconsistent state when there's going something wrong _during_ the initial creation of the PCSs (e.g. if it returns an `err` on line 470 and you have multiple PCS).

controller/stack_resources.go
```diff
@@ -429,3 +429,77 @@ func (c *StackSetController) ReconcileStackSecret(
...
+
+	// TODO:
+	// Check if we need to update a PlatformCredentialsSet
+	for _, e := range existing {
+		if core.IsResourceUpToDate(stack, e.ObjectMeta) {
+			return nil
+		}
+
+		updated := e.DeepCopy()
+		syncObjectMeta(updated, pcs)
```
linki, line 489: `pcs` is the very same PCS in every loop iteration, so this can't be correct. We need to compare `e` to a particular `pcs` in the list of `configurationResources`.

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,36 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
...
+
+	// Ensure that the PlatformCredentialsSet is owned by the Stack
+	suite.Equal([]metav1.OwnerReference{
+		{
+			APIVersion: core.APIVersion,
+			Kind:       core.KindStack,
+			Name:       stack.Name,
+			UID:        stack.UID,
+		},
+	}, pcs.OwnerReferences)
```
katyanna, line 133: TODO: Check `application`, `version`, and `tokens` data.

docs/zalando.org_platformcredentialssets.yaml
```diff
@@ -0,0 +1,108 @@
+---
```
linki, line 1: I think the idea is to _not_ commit this file, according to: https://github.com/zalando-incubator/stackset-controller/pull/614/commits/31e44913274f96a1f00655ffb1c5ed9d975a8830

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,44 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
...
+
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
+		if rsc.IsPlatformCredentialsSet() {
+			suite.Equal(rsc.PlatformCredentialsSet.Tokens, pcs.Spec.Application)
+			suite.Equal(rsc.PlatformCredentialsSet.Tokens, pcs.Spec.TokenVersion)
+			suite.Equal(rsc.PlatformCredentialsSet.Tokens, pcs.Spec.Tokens)
```
linki, line 139: This checks `.Tokens` against different values. I wonder if this can ever pass?
Screenshot 2024-04-10 at 13 28 33

linki:
@katyanna I highly recommend to add the PCS example to the sample deployment files, similar to what I did here: https://github.com/zalando-incubator/stackset-controller/pull/593/files#diff-a5af117a4e2c6ea806cb248dc578041c0a906deb5124378e9dcd67d757d532cf

It's easy to do and allows us to finalize the syntax and it's partially tested by the e2e suite without doing anything.

docs/zalando.org_platformcredentialssets.yaml
katyanna: yep, rebase mistake.

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,43 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
...
+	suite.Equal([]metav1.OwnerReference{
+		{
+			APIVersion: core.APIVersion,
+			Kind:       core.KindStack,
+			Name:       stack.Name,
+			UID:        stack.UID,
+		},
+	}, pcs.OwnerReferences)
+
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
```
linki, line 135: Why do we loop over this? The `rsc` name/data isn't even used in the loop.

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,43 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
 		},
 	}, secret.OwnerReferences)
 }
+
+// TestGeneratedPCS tests that PlatformCredentialsSets defined in the StackSet are
+// correctly created and owned by the Stack.
+func (suite *ConfigurationResourcesTestSuite) TestGeneratedPCS() {
+	// Add the PlatformCredentialsSet reference to the StackSet spec
+	pcsName := "stackset-cr-v1-my-pcs"
+	suite.stacksetSpecFactory.AddPlatformCredentialsSetDefinition(pcsName)
```
linki, line 108: Maybe we allow to pass-in the `PCS` object (similar to [here](https://github.com/zalando-incubator/stackset-controller/pull/596/files#diff-76672fa572a15421d775c154ac0db8c1f5e96b7751659853a6cb7e0b14c24ffaR40-R43)). Then the example is self-contained with respect to what we assert at the end of the test (token name and privileges)

cmd/e2e/configuration_resources_test.go
linki: I think we can just drop the loop, right?

controller/stackset.go
```diff
@@ -313,6 +313,13 @@ func (c *StackSetController) collectResources(ctx context.Context) (map[types.UI
 		}
 	}
 
+	// if c.pcsSupportEnabled {
```
linki, line 316: Let's finalize this by removing the comments.

controller/stackset.go
```diff
@@ -1014,6 +1054,18 @@ func (c *StackSetController) ReconcileStackResources(ctx context.Context, ssc *c
...
```
linki, line 1057: Let's finalize this by removing the comments.

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,43 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
...
+	// Create the StackSet in the cluster
+	err := createStackSet(suite.stacksetName, 0, stacksetSpec)
+	suite.Require().NoError(err)
+
+	// Wait for the first Stack to be created
+	stack, err := waitForStack(suite.T(), suite.stacksetName, suite.stackVersion)
+	suite.Require().NoError(err)
+
+	// Fetch the latest version of the PlatformCredentialsSet
+	pcs, err := waitForPlatformCredentialsSet(suite.T(), pcsName)
```
linki, line 122: This test fails here with an error in stackset-controller:

> stack has no label application

I guess the Stack-under-test needs an application label.

cmd/e2e/configuration_resources_test.go
```diff
@@ -99,3 +99,43 @@ func (suite *ConfigurationResourcesTestSuite) TestReferencedSecrets() {
...
+			Kind:       core.KindStack,
+			Name:       stack.Name,
+			UID:        stack.UID,
+		},
+	}, pcs.OwnerReferences)
+
+	for _, rsc := range stacksetSpec.StackTemplate.Spec.ConfigurationResources {
+		if rsc.IsPlatformCredentialsSet() {
+			suite.Equal("v2", pcs.Spec.TokenVersion)
+			suite.Equal([]string{"read"}, pcs.Spec.Tokens["token-example"].Privileges)
```
linki, line 138: We should also be able to assert for the `application` setting here.

controller/stack_resources_test.go
```diff
@@ -1413,3 +1413,152 @@ func TestReconcileStackSecretRefs(t *testing.T) {
...
+			env := NewTestEnvironment()
+
+			err := env.CreateStacksets(context.Background(), []zv1.StackSet{testStackSet})
+			require.NoError(t, err)
+
+			err = env.CreateStacks(context.Background(), []zv1.Stack{tc.stack})
+			require.NoError(t, err)
+
+			// Create case's existing resources
+			if tc.existing != nil {
```
linki, line 1534: It's very common Golang style to omit this `if` statement. The loop that it guards will just be skipped if `tc.existing` is `nil`.

controller/stack_resources_test.go
```diff
@@ -1413,3 +1413,152 @@ func TestReconcileStackSecretRefs(t *testing.T) {
...
+				for _, existing := range tc.existing {
+					err = env.CreatePCS(
+						context.Background(),
+						[]zv1.PlatformCredentialsSet{*existing},
+					)
+					require.NoError(t, err)
+				}
+			}
```
linki, line 1542: Given that `CreatePCS` already takes a slice of PCS, we may be able to simplify this whole thing to:

```suggestion
			// Create case's existing resources
			err = env.CreatePCS(
				context.Background(),
				tc.existing,
			)
			require.NoError(t, err)
```

We might need to make `CreatePCS` to receive a list of pointers but I think that's a fine compromise.

cmd/e2e/configuration_resources_test.go
```diff
@@ -26,8 +26,13 @@ func (suite *ConfigurationResourcesTestSuite) SetupTest() {
...
 
 	suite.stacksetSpecFactory = NewTestStacksetSpecFactory(suite.stacksetName)
+}
 
-	_ = deleteStackset(suite.stacksetName)
+func (suite *ConfigurationResourcesTestSuite) TearDownTest() {
+	err := deleteStackset(suite.stacksetName)
+	if err != nil {
+		panic(err)
+	}
```
linki, line 35: Let's use `suite.Require().NoError(err)` instead?

cmd/e2e/test_utils.go
```diff
@@ -349,6 +350,12 @@ func createStackSet(stacksetName string, prescalingTimeout int, spec zv1.StackSe
 		Spec:       spec,
 	}
 
+	for _, rsc := range spec.StackTemplate.Spec.ConfigurationResources {
+		if rsc.IsPlatformCredentialsSet() {
+			stackSet.Labels["application"] = stacksetName
```
linki, line 355: I wouldn't do this. It makes the `createStackSet` helper function quite magical because it creates StackSets with the application based on whether it _also_ has a PCS attached. Why not just allow the caller of `createStackSet` to configure the labels by adding a `labels` parameter. Or just add the `application` all the time.

docs/stackset_crd.yaml
```diff
@@ -8063,14 +8059,6 @@ spec:
...
                                                 properties:
                                                   audience:
-                                                    description: audience is the intended
-                                                      audience of the token. A recipient
-                                                      of a token must identify itself
-                                                      with an identifier specified
-                                                      in the audience of the token,
-                                                      and otherwise should reject
-                                                      the token. The audience defaults
-                                                      to the identifier of the apiserver.
```
linki, line 8073: I think these changes are unrelated to this PR, right? Let's just leave them out.

docs/stack_crd.yaml
```diff
@@ -7194,8 +7168,6 @@ spec:
                                               Paths must be relative and may not contain
                                               the '..' path or start with '..'.
                                             items:
-                                              description: Maps a string key to a
-                                                path within a volume.
```
linki, line 7198: I think these changes are unrelated to this PR, right? Let's just leave them out.

katyanna: That's auto trim from make.

docs/stackset_crd.yaml
katyanna: That's auto trim from `make`.

linki: Alright, weird that they aren't trimmed on master 👍 🤔

linki:
👍

katyanna:
👍

PR #588, opened 2024-02-16
Generate HPA pointing to ingress segments.

gargravarr:
The fix in https://github.com/zalando-incubator/stackset-controller/pull/585 didn't work, because when the controller generates HPAs for the first time, the `Resources` in `StackContainer` are empty.

This Pull Requests updates generation of HPAs for a StackSet with enabled segments. I also updated the end-2-end test setup to include an Ingress based metri in tha autoscaler.

### DIFF
--- a/pkg/core/stack_resources_test.go
+++ b/pkg/core/stack_resources_test.go
@@ -1408,34 +1408,23 @@ func TestGenerateHPA(t *testing.T) {
 func TestGenerateHPAToSegment(t *testing.T) {
 	for _, tc := range []struct {
 		name        string
-		minReplicas int32
-		maxReplicas int32
 		metricType  zv1.AutoscalerMetricType
-		metricValue int64
 		expectedRef string
 	}{
 		{
 			name:        "HPA metric points to ingress segment",
-			minReplicas: 1,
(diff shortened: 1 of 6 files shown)


pkg/core/stack_resources_test.go
```diff
@@ -1366,122 +1365,6 @@ func TestGenerateHPA(t *testing.T) {
...
-						Name: v1.ResourceCPU,
-						Target: autoscaling.MetricTarget{
-							Type:               autoscaling.UtilizationMetricType,
-							AverageUtilization: &utilization,
-						},
-					},
-				},
-			},
-			expectedBehavior: exampleBehavior,
-		},
```
mikkeloscar, line 1484: Is it not possible to adjust these test cases to cover the new logic?

gargravarr: I will add tests for `GenerateHPAToSegment`, so we cover that too.

gargravarr: Done

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #592, opened 2024-02-22
Rename functions dealing with config/secret refs for clarify

linki:
For for [follow-up work on inline ConfigMaps](https://github.com/zalando-incubator/stackset-controller/pull/593/files) I've found it helpful to change some functions dealing with the "reference" implementation to include the "Ref" in their names. This way the corresponding inline functions will be different and it's easier to search for them.


mikkeloscar:
:+1:

katyanna:
:+1:

linki:
👍

PR #593, opened 2024-02-22
[2/3] Add Inline Solution for ConfigMaps

linki:
This is a solution for Inline ConfigMaps. It follows the design of other inline resources such as HPAs. It could be used as a blueprint for inline PCS as well.

Please also see the previous and follow up PRs that are based on this one:
- [x] [[1/3] Add test suite dedicated for configuration resources](https://github.com/zalando-incubator/stackset-controller/pull/599)
- [ ] [[3/3] Add e2e test case for Inline ConfigMap](https://github.com/zalando-incubator/stackset-controller/pull/595)

Let's also do this:
- [x] Let's get https://github.com/zalando-incubator/kubernetes-on-aws/pull/7179 merged and run tests in stups-test.


pkg/apis/zalando.org/v1/types.go
```diff
@@ -458,15 +465,30 @@ func (crs *ConfigurationResourcesSpec) GetName() string {
...
+	return crs.ConfigMap != nil && crs.ConfigMap.Name != ""
+}
+
+// IsConfigMapRef returns true if the ConfigurationResourcesSpec is a ConfigMapRef.
 func (crs *ConfigurationResourcesSpec) IsConfigMapRef() bool {
 	return crs.ConfigMapRef != nil && crs.ConfigMapRef.Name != ""
 }
 
-// IsSecret returns true if the ConfigurationResourcesSpec is a Secret.
+// IsSecretRef returns true if the ConfigurationResourcesSpec is a SecretRef.
```
linki, line 477: Needs to be move to the other PR https://github.com/zalando-incubator/stackset-controller/pull/592

controller/stack_resources.go
```diff
@@ -429,3 +429,94 @@ func (c *StackSetController) ReconcileStackSecretRefs(
...
+	for _, existingConfigMap := range existingConfigMaps {
+		desiredConfigMap := findConfigMap(desiredConfigMaps, existingConfigMap.Name)
+		if desiredConfigMap == nil {
+			if err := c.ReconcileStackConfigMap(ctx, stack, existingConfigMap, desiredConfigMap); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
```
linki, line 456: This part is necessary to ensure that stackset-controller really "reconciles", i.e. if you add/remove a configurationResource from a Stack, the ConfigMap will be created/deleted accordingly. If you change the inline template it would be updated. Similarly if you delete a generated ConfigMap it would be recreated.

controller/stack_resources.go
```diff
@@ -429,3 +429,94 @@ func (c *StackSetController) ReconcileStackSecret(
...
+		c.recorder.Eventf(
+			stack,
+			apiv1.EventTypeNormal,
+			"CreatedConfigMap",
+			"Created ConfigMap %s",
+			desiredConfigMap.Name)
+		return nil
+	}
+
+	// Check if we need to update the RouteGroup
```
linki, line 493: typo: RouteGroup

pkg/core/stack_resources.go
```diff
@@ -571,6 +571,36 @@ func (sc *StackContainer) UpdateObjectMeta(objMeta *metav1.ObjectMeta) *metav1.O
...
+				APIVersion: "zalando.org/v1",
+				Kind:       "Stack",
+				Name:       sc.Name(),
+				UID:        sc.Stack.UID,
+			},
+		}
+
+		base := &v1.ConfigMap{
+			ObjectMeta: metaObj,
+			Data:       configResource.ConfigMap.Data,
```
linki, line 595: I think for the (original) reference approach, we could just do a
```go
client.CoreV1().ConfigMaps(metaObj.Namespace).Get(ctx, configResource.GetName(), metav1.GetOptions{})
```
to fetch the reference ConfigMap's data and other fields. We'd need to pass-in the Kubernetes client for that but everything else would stay the same.

controller/stack_resources.go
```diff
@@ -472,3 +472,94 @@ func (c *StackSetController) ReconcileStackSecretRef(ctx context.Context,
...
+
+	return nil
+}
+
+func (c *StackSetController) ReconcileStackConfigMap(ctx context.Context, stack *zv1.Stack, existingConfigMap *apiv1.ConfigMap, desiredConfigMap *apiv1.ConfigMap,
+) error {
+	// ConfigMap removed
+	if desiredConfigMap == nil {
+		if existingConfigMap != nil {
+			err := c.client.CoreV1().ConfigMaps(existingConfigMap.Namespace).Delete(ctx, existingConfigMap.Name, metav1.DeleteOptions{})
```
linki, line 507: This doesn't handle the case where a ConfigMap with the same name already exists in the cluster but isn't defined in the StackSet definition as a `configurationResource`. In that case the resource would be deleted. I think the same would happen for HPAs, Ingress, RouteGroups etc. but maybe we want to do an owner check similar to how we already do it for reference ConfigMaps (where we don't touch the resource if it's not owned by us).

PR #595, opened 2024-02-23
[3/3] Add e2e test case for Inline ConfigMaps

linki:
This adds a first e2e test for Inline ConfigMaps developed in https://github.com/zalando-incubator/stackset-controller/pull/593. It's based on the dedicated test suite created in https://github.com/zalando-incubator/stackset-controller/pull/599.

It handles a very simple case where an inline ConfigMap with a single key-value pair of data is defined. It validates that a ConfigMap is created in the cluster with the expanded name (including the StackSet name and Stack version) as well as that it's owned by the Stack via ownerReferences.

TODOs

The rewrite portion hasn't been implemented yet, so the tests for that are skipped (see comment). I'd suggest to start by adding a test case here because the rewrite is probably the most delicate problem of this feature.

Please also see the previous PRs that are needed for this one:
- [x] [[1/3] Add test suite dedicated for configuration resources](https://github.com/zalando-incubator/stackset-controller/pull/599)
- [ ] [[2/3] Add Inline Solution for ConfigMaps](https://github.com/zalando-incubator/stackset-controller/pull/593)


linki:
👍

linki:
Closing this as inactive. Let's follow up if we decide to support inline ConfigMaps. (This branch is supposed to be merged into `configmap-inline-another-try` branch from the other PR)

PR #599, opened 2024-02-29
[1/3] Add test suite dedicated for configuration resources

linki:
Separate test suite to focus on versioned configuration resources where we don't need to care (as much) about all that other stuff that StackSets can have.

The rewrite of the existing tests using the new methods to make sure everything still works was done in https://github.com/zalando-incubator/stackset-controller/pull/606. Some of these could be removed as the behaviour is tested in this new suite as well but we can also leave them.

Please also see the follow up PRs that are based on this one:
- [[2/3] Add Inline Solution for ConfigMaps](https://github.com/zalando-incubator/stackset-controller/pull/593)
- [[3/3] Add e2e test case for Inline ConfigMap](https://github.com/zalando-incubator/stackset-controller/pull/595)

### DIFF
--- a/cmd/e2e/configuration_resources_test.go
+++ b/cmd/e2e/configuration_resources_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
@@ -26,10 +25,9 @@ func (suite *ConfigurationResourcesTestSuite) SetupTest() {
 	suite.stacksetName = "stackset-cr"
 	suite.stackVersion = "v1"
 
-	// TODO: The cleanup should be done automatically but without being too magical
-	suite.cleanUpStackSet(suite.stacksetName)
-
 	suite.stacksetSpecFactory = NewTestStacksetSpecFactory(suite.stacksetName)
+
+	_ = deleteStackset(suite.stacksetName)
 }
 
 // TestReferencedConfigMaps tests that ConfigMaps referenced in the StackSet spec are owned by the Stack.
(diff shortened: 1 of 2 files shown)


cmd/e2e/configuration_resources_test.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+	}, secret.OwnerReferences)
+}
+
+//
+// HELPER FUNCTIONS
+//
+
+// cleanUpStackSet deletes the StackSet with the given name.
+func (suite *ConfigurationResourcesTestSuite) cleanUpStackSet(name string) {
```
gargravarr, line 110: I added a similar function for Stacks here: https://github.com/zalando-incubator/stackset-controller/blob/master/cmd/e2e/test_utils.go#L386-L392

So maybe move this function, which we can use in other test cases.

linki: Will do 👍 

I honestly didn't check `test_utils.go` for an existing function because I wanted to look at it with a fresh set of eyes. I'm wondering how we can structure it in a way where it's easy to share these functions between test cases (which we should) while at the same time keep it organised. By organised I mean to have a set of meaningful helper functions that we can effectively find and re-use as well as add more functions in a similar manner. Just putting everything in `test_utils.go` has its limits.

gargravarr: I agree with that and am in principle agains any "utils" package / file. The sweet spot here is if it is worth doing a larger refarctoring: something like a generic `TestSuite` with proper cleanup.

linki: Resolved for now by adding a `deleteStackset` function similar to `deleteStack` in `test_utils`.

linki:
👍

PR #605, opened 2024-03-01
Add a versioned configmap to the e2e sample stackset

linki:
This uses a versioned ConfigMap in the sample deployment so we have this covered.


linki:
👍

katyanna:
👍

PR #606, opened 2024-03-01
Allow more control over configured versioned resources for more flexible test cases

linki:
This came out while working on https://github.com/zalando-incubator/stackset-controller/pull/599

This replaces the specific fields `configMapRef` and `secretRef` that were used to attach a versioned resource to a StackSet with the more granular functions:
* `AddReferencedConfigMap`/`AddReferencedSecret` which add an entry for the given resource under StackSet's `configurationResources` section
* `volumes` which hold the configured volumes for the StackSet. (There's no dedicated accessor function for volumes because it's not needed here.)

Both functions can be used in a straight-forward way to build-up a StackSet into a particular state for testing, i.e. having certain versioned resources in this case.

The author of the test case is still required to **actually create** the referenced versioned resource similar to how users would do it via CDP (see `createConfigMap` and `createSecret` helper functions for that). This is to keep the test close to reality and allow for testing certain failure scenarios, for example when a reference is defined in the StackSet but the referenced resource doesn't exist which can easily happen in reality. This is obviously debatable as it _yet again_ makes writing test cases more error-prone but I didn't want to throw too many things together at this stage.

### DIFF
--- a/cmd/e2e/basic_test.go
+++ b/cmd/e2e/basic_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"slices"
 	"sort"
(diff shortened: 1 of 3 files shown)


gargravarr:
:+1:

PR #608, opened 2024-03-12
Update k8s.io/* dependencies to v0.26.14

zaklawrencea:
This PR updates:
- k8s.io/* to v0.26.14
- generated code
- additional dependencies


controller/stackset.go
```diff
@@ -634,7 +634,7 @@ func (c *StackSetController) startWatch(ctx context.Context) {
 		cache.Indexers{},
 	)
 
-	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+	_, _ = informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
gargravarr, line 637: why the anonymous variables?

mikkeloscar: Yeah, would make sense we handle the return values especially as this seem to be new in v1.26?

zaklawrencea: I updated this due to golangci-lint. However I didn't want to edit more than strictly necessary as this is primarily a dependency update.

```bash
golangci-lint run --timeout=2m ./...
Error: controller/stackset.go:637:26: Error return value of `informer.AddEventHandler` is not checked (errcheck)
	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
	                        ^
make: *** [Makefile:32: check] Error 1
```

mikkeloscar: But this is only triggering _after_ the dependency update right? Or was it also failing prior to updating?

gargravarr: This is when it could return an error. Highly unlikely, but doesn't hurt to check for the error:

https://github.com/kubernetes/client-go/blob/62c5e9e7b4bd09d2b43dcad1464376322398e994/tools/cache/shared_informer.go#L579-L581

zaklawrencea: > But this is only triggering after the dependency update right?

Yes, correct. This is only after the updates.

I'm happy to add a check for the error. Do you have a preference for handling it in this case?

linki: Just return it an bubble it up.

controller/stackset.go
```diff
@@ -626,25 +629,30 @@ func (c *StackSetController) hasOwnership(stackset *zv1.StackSet) bool {
...
 	)
 
-	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    c.add,
 		UpdateFunc: c.update,
 		DeleteFunc: c.del,
 	})
+	if err != nil {
+		return fmt.Errorf("Failed to add event handler: %w", err)
```
mikkeloscar, line 646: nit: errors should not be capitalized: https://go.dev/wiki/CodeReviewComments#:~:text=Error%20strings%20should%20not%20be,something%20bad%22)%20not%20fmt.

(Same for the error below on 651.

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #610, opened 2024-03-15
Fix controller-tools crash on Go 1.22

mikkeloscar:
This is a fix for the bug described here: https://github.com/golang/go/issues/65637 where the controller-tools could crash when run with Go 1.22

The fix is basically to update `golang.org/x/tools` to `v0.19.0` as outlined here: https://github.com/golang/go/issues/65637#issuecomment-1998639426


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #614, opened 2024-03-27
Use non-anonymous types for PCS status

mikkeloscar:
Follow up to #580 and specifically https://github.com/zalando-incubator/stackset-controller/pull/580#discussion_r1540968233

* Drop the `/bin/true || true` hack in CRD generation
* Use defined types for clients/tokens in PCS status. This is not really important for how this is used in stackset-controller, but makes the CRD generation work without errors and is more in line with the actual status.


mikkeloscar:
:+1:

AlexanderYastrebov:
Nice. 
BTW do we need to generate PCS CRD at all? And do we need PCS status?

mikkeloscar:
> BTW do we need to generate PCS CRD at all? And do we need PCS status?

Not really, we only have it for creating/reading the PCS, the CRD we don't need, didn't check if it's simple to exclude it from generation, but we can ofc. do that.

AlexanderYastrebov:
> BTW do we need to generate PCS CRD at all? 

Looks like it is only possible on package level via
```go
// +kubebuilder:skip
```

> And do we need PCS status?

My idea was that if we only need to read/embed PCS spec then we may just drop PCS status fields and related types.
We may also add `rm -f docs/zalando.org_platformcredentialssets.yaml` to the make target.

mikkeloscar:
I have added `rm -f docs/zalando.org_platformcredentialssets.yaml`.

AlexanderYastrebov:
:+1:

PR #618, opened 2024-04-04
Remove End-to-end testing for legacy central ingress resources.

gargravarr:
Removes End-2-End tests of cases using central ingress resources. We don't need to test these cases  anymore, because from now StackSets use only traffic ingress segment resources. 

* Removes all central ingress test cases from basic tests.
* Adapts tests relying on central ingress resources to verify instead ingress segment resources.
* Removes test cases tesing switching a StackSet from having an Ingress to a RouteGroup and vice-versa. With segments we don't have this problem, because each stack has its own ingress resource, whether is an Ingress, RouteGroup, or both.
* Removes helper functions to verify central ingress resources.
* Modifies end-to-end setup to use only ingress versioning.

We need to have a version of the StackSet Controller without the above tests before removing the Ingress versioning config items from kubernetes-on-aws. A follow-up Pull Request will remove the logic for central Ingress resources.


katyanna:
Should `verifyStackIngressSources` on `cmd/e2e/basic_test.go` also be removed?

There's also some `waitForIngress()` still being used `ttl_test.go` and `prescalling_test.go`, could we remove that too?

gargravarr:
> Should `verifyStackIngressSources` on `cmd/e2e/basic_test.go` also be removed?
> 
We can't because the Stacks themselves have their own Ingress named `stack-name-version`, besides the IngressSegment, we use it here for example: https://github.com/zalando-incubator/stackset-controller/blob/b61c457eb700a75d6ec43f9e36637bee340084ec/cmd/e2e/basic_test.go#L332-L338

> There's also some `waitForIngress()` still being used `ttl_test.go` and `prescalling_test.go`, could we remove that too?

The same reason as above, the test is waiting for a Stack's ingress, not the IngressSegment: https://github.com/zalando-incubator/stackset-controller/blob/b61c457eb700a75d6ec43f9e36637bee340084ec/cmd/e2e/basic_test.go#L464

katyanna:
:+1:

gargravarr:
:+1:

PR #621, opened 2024-04-11
Add inline configmaps to the stackset spec

linki:
Part of https://github.com/zalando-incubator/stackset-controller/pull/593 (just the CRD spec)

As pointed out in https://github.com/zalando-incubator/kubernetes-on-aws/pull/7179#pullrequestreview-1994123258 the optional (but merged) Stackset definition hasn't been pushed to master here. This might lead to confusion. Although https://github.com/zalando-incubator/stackset-controller/pull/593 is not yet finished, let's already merge the spec so that there aren't any merge conflicts.


linki:
👍

PR #622, opened 2024-04-12
Add platformCredentialsSets to the supported configurationResources

linki:
This adds `platformCredentialsSets` to the supported `configurationResources` and validates it by adding an entry in the sample.

Note, there's no functionality yet behind the new field in the CRD.

- [x] Depends on https://github.com/zalando-incubator/kubernetes-on-aws/pull/7304 to be enabled in `stups-test`.


e2e/apply/sample.yaml
```diff
@@ -23,6 +23,12 @@ spec:
           name: e2e-deploy-sample-{{{CDP_BUILD_VERSION}}}-my-reference-config
       - secretRef:
           name: e2e-deploy-sample-{{{CDP_BUILD_VERSION}}}-my-reference-secret
+      - platformCredentialsSet:
+          name: e2e-deploy-sample-{{{CDP_BUILD_VERSION}}}-my-pcs
+          tokens:
+            sample-token:
+              privileges:
+              - com.zalando::foobar.read
```
linki, line 31: Defining it here is just a quick way to test that our spec is correct.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -449,6 +449,16 @@ type ConfigurationResourcesSpec struct {
...
 	// SecretRef is a reference to a Secret to be owned by Stack
 	SecretRef *v1.LocalObjectReference `json:"secretRef,omitempty"`
+
+	// PlatformCredentialsSet to be created and owned by Stack
+	PlatformCredentialsSet *PCS `json:"platformCredentialsSet,omitempty"`
+}
+
+// PCS is the PlatformCredentialsSet definition for a stack
+// +k8s:deepcopy-gen=true
+type PCS struct {
```
linki, line 459: @katyanna Let's rename this to something more meaningful. Looking at the structure of this and the "real" `PlatformCredentialsSet` this looks like a condensed for of it (no extra `spec` layer) and it only contains `Tokens` (no clients, token version, application). We should find a proper name for that otherwise it's easy to think that this is a "full" PCS.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -483,6 +497,11 @@ func (crs *ConfigurationResourcesSpec) IsSecretRef() bool {
 	return crs.SecretRef != nil && crs.SecretRef.Name != ""
 }
 
+// IsPlatformCredentialsSet returns true if the ConfigurationResourcesSpec is an inline PlatformCredentialsSet.
+func (crs *ConfigurationResourcesSpec) IsPlatformCredentialsSet() bool {
+	return crs.PlatformCredentialsSet != nil && crs.PlatformCredentialsSet.Name != ""
```
linki, line 502: @katyanna fyi, I added `crs.PlatformCredentialsSet.Name != ""` here similar to the other functions.

katyanna:
:+1:

linki:
👍

PR #591, opened 2024-02-22
Fix order of metrics in autoscaling test

mikkeloscar:
When generating an HPA from the `autoscaler` config we sort the metrics based on [some heuristic](https://github.com/zalando-incubator/stackset-controller/blob/5e8d42ca5d32ecfe938fed43e8a19de527254f07/pkg/core/autoscaler.go#L60-L63) to have a consistent naming for metrics.

In the e2e test the order in which we check the metrics does not match.

~This issue only shows up in e2e tests in Kubernetes v1.28 which I don't quite understand, as I would think it would be broken in earlier versions as well :shrug: This should be a general fix none the less.~ This is an issue with Kubernetes below v1.28 See https://github.com/zalando-incubator/stackset-controller/pull/591#issuecomment-1959751276

PR #624, opened 2024-04-19
Check for expected order of HPA metrics

mikkeloscar:
in #591 we introduced a mitigation for HPA metrics not being in the submitted order in Kubernetes <v1.28. Now that we're on v1.28 or above, we can expect the metrics to be in the submitted order. Therefore adjust the test case to check for the order.


mikkeloscar:
:+1:

linki:
👍

PR #626, opened 2024-04-24
Update production image base image

mikkeloscar:
* Update the base image used for the tagged image to be based on `alpine-3` instead of the old version `alpine-3.13`
* Also update `golang.org/x/net` to latest version.


mikkeloscar:
:+1:

katyanna:
👍

PR #627, opened 2024-04-29
Wait for ownerReferences to appear on the resources

linki:
For https://github.bus.zalan.do/teapot/issues/issues/3622

We check referenced versioned ConfigMaps and Secrets for their ownerReferences. However, we need to allow `stackset-controller` a little bit of time to process them. Typically resources are _created_ by `stackset-controller` so that upon waiting for them they are already in their final state. This is not the case for referenced configuration resources which exist prior to processing, so we need to delay the check a bit. Note that inline PlatformCredentialsSets don't need it because they are created by `stackset-controller`.

### DIFF
--- a/cmd/e2e/test_utils.go
+++ b/cmd/e2e/test_utils.go
@@ -486,12 +486,38 @@ func waitForConfigMap(t *testing.T, configMapName string) (*corev1.ConfigMap, er
 	return configMapInterface().Get(context.Background(), configMapName, metav1.GetOptions{})
 }
 
+func waitForConfigMapOwnerReferences(t *testing.T, configMapName string, expectedOwnerReferences []metav1.OwnerReference) *awaiter {
+	return newAwaiter(t, fmt.Sprintf("configmap %s to reach desired ownerReferences", configMapName)).withPoll(func() (retry bool, err error) {
+		configMap, err := configMapInterface().Get(context.Background(), configMapName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		if !reflect.DeepEqual(expectedOwnerReferences, configMap.OwnerReferences) {
+			return true, fmt.Errorf("%s: actual ownerReferences: %+v, expected: %+v, diff:\n%v", configMapName, configMap.OwnerReferences, expectedOwnerReferences, cmp.Diff(configMap.OwnerReferences, expectedOwnerReferences))
+		}
+		return true, nil
+	})
+}
+
 func waitForSecret(t *testing.T, secretName string) (*corev1.Secret, error) {
 	err := resourceCreated(t, "secret", secretName, secretInterface()).await()
 	if err != nil {
 		return nil, err
 	}
 	return secretInterface().Get(context.Background(), secretName, metav1.GetOptions{})
+}
+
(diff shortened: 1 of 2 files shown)


mikkeloscar:
@linki I don't think 1 second is enough. What if we're unlucky and the stackset controller pod is being rescheduled right at this moment in time?

linki:
Right, so something like a `waitForOwnerReference` after the `waitForConfigMap`? That should cover all the cases.

mikkeloscar:
@linki Yeah, seems reasonable!

linki:
👍

cmd/e2e/test_utils.go
```diff
@@ -486,6 +486,19 @@ func waitForConfigMap(t *testing.T, configMapName string) (*corev1.ConfigMap, er
...
+func waitForConfigMapOwnerReferences(t *testing.T, configMapName string, expectedOwnerReferences []metav1.OwnerReference) *awaiter {
+	return newAwaiter(t, fmt.Sprintf("configmap %s to reach desired ownerReferences", configMapName)).withPoll(func() (retry bool, err error) {
+		configMap, err := configMapInterface().Get(context.Background(), configMapName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		if !reflect.DeepEqual(expectedOwnerReferences, configMap.OwnerReferences) {
+			return true, fmt.Errorf("%s: actual ownerReferences: %+v, expected: %+v, diff:\n%v", configMapName, configMap.OwnerReferences, expectedOwnerReferences, cmp.Diff(configMap.OwnerReferences, expectedOwnerReferences))
+		}
+		return true, nil
```
mikkeloscar, line 498: I think this should be `false` such that it doesn't retry if it was successful?

linki: Yes 🤦

I checked and realised that the value is ignored _if_ the err is nil (so you cannot retry if there's no error). That's why it worked.

cmd/e2e/test_utils.go
```diff
@@ -494,6 +507,19 @@ func waitForSecret(t *testing.T, secretName string) (*corev1.Secret, error) {
...
+func waitForSecretOwnerReferences(t *testing.T, secretName string, expectedOwnerReferences []metav1.OwnerReference) *awaiter {
+	return newAwaiter(t, fmt.Sprintf("secret %s to reach desired ownerReferences", secretName)).withPoll(func() (retry bool, err error) {
+		secret, err := secretInterface().Get(context.Background(), secretName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		if !reflect.DeepEqual(expectedOwnerReferences, secret.OwnerReferences) {
+			return true, fmt.Errorf("%s: actual ownerReferences: %+v, expected: %+v, diff:\n%v", secretName, secret.OwnerReferences, expectedOwnerReferences, cmp.Diff(secret.OwnerReferences, expectedOwnerReferences))
+		}
+		return true, nil
```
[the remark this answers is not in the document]
linki: Fixed

linki:
👍

mikkeloscar:
:+1:

PR #635, opened 2024-05-30
Use different stackset names for parallel test cases

linki:
I though they didn't but apparently these tests run in parallel causing conflicts during test runs.

I kept the suite setup for now becuase of the handy `Setup` and `TearDown` functions which would otherwise have to be inlined.


linki:
👍

PR #665, opened 2024-12-11
Use a valid app ID for PCS in e2e test

mikkeloscar:
Use a valid appID for PCS e2e test:

image

### DIFF
--- a/e2e/apply/sample.yaml
+++ b/e2e/apply/sample.yaml
@@ -3,7 +3,8 @@ kind: StackSet
 metadata:
   name: e2e-deploy-sample
   labels:
-    application: "e2e-deploy-sample"
+    application: "kubernetes"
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

e2e/apply/sample.yaml
```diff
@@ -38,9 +38,6 @@ spec:
         - type: Ingress
           average: 20000m
       podTemplate:
-        metadata:
-          labels:
-            application: "e2e-deploy-sample"
```
demonCoder95, line 43: Is it not important to have the `application: kubernetes` label here as well?

demonCoder95: Ah, because of the parenthood. Got it. Thanks.

linki:
👍

PR #690, opened 2025-06-20
shorten subdomain name in stackset hostnames

demonCoder95:
This is needed to conform to subdomain length limit of `57` characters introduced recently that is causing e2e tests to break up. 

The namespace prefix like `d-9owpufu1wutzdgxfdrpytgzw1-` comprises `28` characters which occupy half of this limit. So, splitting the namespace into a subdomain of its own restores the limit to `57` for the stack name alone. This is sufficient to cover the stackset names which can be very long in some tests.


demonCoder95:
> the DNS names are not used as part of the e2e test so it's probably fine

Let's merge then!

demonCoder95:
👍

linki:
👍

PR #693, opened 2025-06-24
Make per-stack hostnames configurable

demonCoder95:
This makes per-stack hostnames configurable in the stackset. 

This allows us to disable support for per-stack hostnames, which is required to solve the problem of duplication of hostnames when a stackset is deployed in multiple namespaces, or multiple clusters in the same account.

The objective is to first be able to disable the support, and then encourage the users that actually rely on this to use `.cluster.local` hostnames instead for their e2e testing.

### DIFF
--- a/cmd/stackset-controller/main.go
+++ b/cmd/stackset-controller/main.go
@@ -71,7 +71,7 @@ func main() {
 	kingpin.Flag("enable-configmap-support", "Enable support for ConfigMaps on StackSets.").Default("false").BoolVar(&config.ConfigMapSupportEnabled)
 	kingpin.Flag("enable-secret-support", "Enable support for Secrets on StackSets.").Default("false").BoolVar(&config.SecretSupportEnabled)
 	kingpin.Flag("enable-pcs-support", "Enable support for PlatformCredentialsSet on StackSets.").Default("false").BoolVar(&config.PCSSupportEnabled)
-	kingpin.Flag("enable-per-stack-hostname", "Enable support for per-stack hostnames.").Default("false").BoolVar(&config.PerStackHostnameEnabled)
+	kingpin.Flag("enable-per-stack-hostname", "Enable support for per-stack hostnames.").Default("true").BoolVar(&config.PerStackHostnameEnabled)
 	kingpin.Parse()
 
 	if config.Debug {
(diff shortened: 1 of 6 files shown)


pkg/core/stack_resources_test.go
```diff
@@ -570,7 +570,7 @@ func TestGenerateIngressSegmentWithSyncAnnotations(t *testing.T) {
 			ingressAnnotationsToSync: tc.ingresssAnnotationsToSync,
 			syncAnnotationsInIngress: tc.syncAnnotationsInIngress,
 		}
-		res, _ := c.GenerateIngressSegment()
+		res, _ := c.GenerateIngressSegment(true)
```
linki, line 573: Can we duplicate this block of tests for `GenerateIngressSegment(false)`?

demonCoder95:
Will test e2e tests by setting default to `true`, which should make the flag a no-op.

cmd/stackset-controller/main.go
```diff
@@ -70,6 +71,7 @@ func main() {
 	kingpin.Flag("enable-configmap-support", "Enable support for ConfigMaps on StackSets.").Default("false").BoolVar(&config.ConfigMapSupportEnabled)
 	kingpin.Flag("enable-secret-support", "Enable support for Secrets on StackSets.").Default("false").BoolVar(&config.SecretSupportEnabled)
 	kingpin.Flag("enable-pcs-support", "Enable support for PlatformCredentialsSet on StackSets.").Default("false").BoolVar(&config.PCSSupportEnabled)
+	kingpin.Flag("enable-per-stack-hostname", "Enable support for per-stack hostnames.").Default("false").BoolVar(&config.PerStackHostnameEnabled)
```
mikkeloscar, line 74: nit: Let's make it true by default to be backwards comaptible.

pkg/core/stack_resources.go
```diff
@@ -379,11 +379,15 @@ func (sc *StackContainer) GenerateIngressSegment() (
...
 }
 
-func (sc *StackContainer) generateIngress(segment bool) (
+func (sc *StackContainer) generateIngress(segment bool, perStackHostnameEnabled bool) (
 	*networking.Ingress,
 	error,
 ) {
 
+	// If Per-stack Hostnames are disabled, no need to generate per stack ingresses
+	if !perStackHostnameEnabled {
```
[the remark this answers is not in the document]
demonCoder95: Ah, I see. I'll try with that. Thanks.

demonCoder95:
Closing this in favor of #696 which implements a better solution.

PR #702, opened 2025-07-14
Make internal cluster domain optional during e2e runs

linki:
The idea is to be able to run e2e tests with only one domain.


demonCoder95:
👍

PR #704, opened 2025-07-21
adapt traffic command to the possibility of missing non-segment routegroup/ingress

linki:
Follow up on https://github.com/zalando-incubator/kubernetes-on-aws/pull/9677

When public domains are filtered out on Stacks, like in the PR above, it can happen that individual Stacks don't have any RouteGroup or Ingress (besides the traffic segment). This trips up the traffic switch command as it waits for both the legacy ingress/routegroups as well as the traffic segments to become ready. The former doesn't exist so it never happens and hangs.

This change makes `stackset-controller` consider the RouteGroup "ready" when it wouldn't have any hostnames matching any cluster domains anyway. The respective code for not creating the RouteGroup/Ingress in this case is [here](https://github.com/zalando-incubator/stackset-controller/blob/735edd31a7f352a9cfc5b11ea07908d6886916b1/pkg/core/stack_resources.go#L489). 

Traffic switching these days works entirely via the Traffic Segment RouteGroup/Ingresses, so maybe the old ones shouldn't be considered _at all_ anymore. This is one step in this direction where it ignores them if it doesn't create them in the first place.

### DIFF
--- a/pkg/core/types.go
+++ b/pkg/core/types.go
@@ -441,9 +441,12 @@ func (sc *StackContainer) updateFromResources() {
 	if sc.ingressSpec != nil {
 		// the per-stack ingress must either be present and up-to-date, or not present and not expected.
 		// the per-stack ingress is not expected if the stack has no hostnames matching the cluster domain.
-		hostnames := sc.stackHostnames(sc.ingressSpec, false)
-		ingressNotExpected := len(hostnames) == 0
-		ingressUpdated = sc.Resources.Ingress == nil && ingressNotExpected || sc.Resources.Ingress != nil && IsResourceUpToDate(sc.Stack, sc.Resources.Ingress.ObjectMeta)
+		if sc.Resources.Ingress != nil {
+			ingressUpdated = IsResourceUpToDate(sc.Stack, sc.Resources.Ingress.ObjectMeta)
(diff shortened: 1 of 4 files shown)


pkg/core/types.go
```diff
@@ -439,18 +439,20 @@ func (sc *StackContainer) updateFromResources() {
 
 	// ingress: ignore if ingress is not set or check if we are up to date
 	if sc.ingressSpec != nil {
-		ingressUpdated = sc.Resources.Ingress != nil && IsResourceUpToDate(sc.Stack, sc.Resources.Ingress.ObjectMeta)
+		hostnames := sc.stackHostnames(sc.ingressSpec, false)
+		ingressUpdated = len(hostnames) == 0 || sc.Resources.Ingress != nil && IsResourceUpToDate(sc.Stack, sc.Resources.Ingress.ObjectMeta)
```
demonCoder95, line 443: It would be nice to add a comment above this line to explain why the `len(hostnames) == 0` is here. The solution makes sense right now but might be too opaque for someone trying to understand it later.

Same comment for the routegroup code below.

demonCoder95: Looks good!

linki:
👍

PR #705, opened 2025-07-22
Add routegroup to test cases to make them useful

linki:
The pattern of these test cases is to check for

```go
require.EqualValues(t, false, container.resourcesUpdated)
```

which just means: "assert that any resource hasn't updated"

Several test cases pass this test _but_ for the wrong reason. For example, the Ingress should not be considered updated because its generation isn't the latest, hence the test passes. However, even if you change it to the latest generation the test will still pass. This is because [_some_ other resource is also not updated](https://github.com/zalando-incubator/stackset-controller/blob/735edd31a7f352a9cfc5b11ea07908d6886916b1/pkg/core/types.go#L473-L479), making the test pass. But clearly, this isn't desired because it makes the test pretty much useless (if you break the ingress functionality, the test will still pass).

This just tests ingress, routegroup and segments individually to surface whether each of them works correctly. This surfaced a small bug which is fixed as well.

### DIFF
--- a/pkg/core/stackset_test.go
+++ b/pkg/core/stackset_test.go
@@ -868,6 +868,14 @@ func TestStackUpdateFromResources(t *testing.T) {
 		container.updateFromResources()
 		require.EqualValues(t, false, container.resourcesUpdated)
 	})
+	runTest("ingress isn't considered updated if it should be gone", func(t *testing.T, container *StackContainer) {
+		container.Stack.Generation = 11
+		container.Resources.Deployment = deployment(11, 5, 5)
+		container.Resources.Service = service(11)
+		container.Resources.Ingress = ingress(11)
+		container.updateFromResources()
+		require.EqualValues(t, false, container.resourcesUpdated)
+	})
 	runTest("ingress segment isn't considered updated if the generation is different", func(t *testing.T, container *StackContainer) {
 		container.Stack.Generation = 11
 		container.ingressSpec = &zv1.StackSetIngressSpec{}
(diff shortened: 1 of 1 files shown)


pkg/core/stackset_test.go
```diff
@@ -866,6 +875,57 @@ func TestStackUpdateFromResources(t *testing.T) {
...
+	runTest("ingress segment isn't considered updated if the generation is different", func(t *testing.T, container *StackContainer) {
+		container.Stack.Generation = 11
+		container.ingressSpec = &zv1.StackSetIngressSpec{}
+		container.Resources.Deployment = deployment(11, 5, 5)
+		container.Resources.Service = service(11)
+		container.Resources.IngressSegment = ingress(10)
+		container.updateFromResources()
+		require.EqualValues(t, false, container.resourcesUpdated)
+	})
+	runTest("ingress segment isn't considered updated if it should be gone", func(t *testing.T, container *StackContainer) {
```
demonCoder95, line 887: Can you explain the "it should be gone" part? I can't read it from the test case code, what part exactly makes the ingress segment "gone"?

demonCoder95: Would be nice to add a comment on the test case so it makes it easier to understand the test cases in the future :)

linki: > Oh, is it because the spec hasn't been defined on the stackset container? I think that's it.

Exactly, it's the absence of `container.ingressSpec = &zv1.StackSetIngressSpec{}`.

This comes from the other test cases, I just took this approach from there. I wouldn't add a comment just for this test case, to be honest.

demonCoder95:
👍

PR #737, opened 2026-06-05
feat(autoscaler): add Prometheus metric type (e.g. for GPU autoscaling)

elgalu:
# Add a generic `Prometheus` autoscaler metric type

## What

Adds `Prometheus` to the StackSet `autoscaler.metrics[].type` enum. A `Prometheus` metric carries a
free-form PromQL `query` and compiles to a Kubernetes HPA `External` metric with a `type: prometheus`
selector plus a `metric-config.external.<name>.prometheus/query` annotation. That annotation is the
documented contract for [kube-metrics-adapter](https://github.com/zalando-incubator/kube-metrics-adapter)'s
Prometheus collector, which runs the query and serves the value to the HPA. No new controller or
adapter code is needed to serve the metric.

## Why

Any metric already in Prometheus becomes an autoscaling signal without a new metric type per source.
The motivating use case is GPU autoscaling for NVIDIA Triton on EKS (scaling on `DCGM_FI_DEV_GPU_UTIL`
or Triton queue-depth metrics), but the primitive is deliberately generic. A single free-form PromQL
type covers GPU and every other Prometheus-resident metric, rather than baking in a niche GPU type.

## Usage

```yaml
autoscaler:
  minReplicas: 1
  maxReplicas: 8
  metrics:
  - type: Prometheus
    average: "70"
    prometheus:
      query: scalar(avg(DCGM_FI_DEV_GPU_UTIL{container="triton"}))
```

A stack may declare several `Prometheus` metrics; each gets a unique name (`<stack>-prometheus-<i>`),
so queries and annotations do not collide. The query must return a scalar or single-element vector.

## Testing

- `go build ./...`, `go vet`, `go test ./pkg/...` all pass (two new tests plus existing sort test).
- Verified end-to-end on a EKS test cluster against `kube-metrics-adapter v0.2.9-6-gf2c845a`:
  an HPA with the generated metric shape (`type: prometheus` selector + query annotation) was served
  by the adapter and scaled correctly. Confirmed the selector label and annotation key match the
  deployed adapter version.


elgalu:
Conceptual work, we'll avoid GPU-based scaling for now.

PR #755, opened 2026-09-03
Make forward compatible with pre-scale

mikkeloscar:
This fixes a bug where forward (#716, #721)  in combination with `alpha.stackset-controller.zalando.org/prescale-stacks` would not work and switching traffic to a forwarded stack would get stuck because the "virtual" stack was not ready.

The solution is to not check for readiness of the underlying deployment, as there is no underlying deployment when the stack is marked for forwarding.

### DIFF
--- a/pkg/core/test_helpers.go
+++ b/pkg/core/test_helpers.go
@@ -99,6 +99,14 @@ func (f *testStackFactory) pendingRemoval() *testStackFactory {
 	return f
 }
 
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

szuecs (approved):

szuecs:
:+1:

szuecs:
I lost my write access, for some reason

mikkeloscar:
> I lost my write access, for some reason

I added you now!

